PR #2, opened 2025-03-05
feat: implement rokt launcher into initForwarder

crisryantan:
## Instructions
 1. PR target branch should be against `development`
 2. PR title name should follow this format: https://github.com/mParticle/mparticle-workflows/blob/main/.github/workflows/pr-title-check.yml
 3. PR branch prefix should follow this format: https://github.com/mParticle/mparticle-workflows/blob/main/.github/workflows/pr-branch-check-name.yml

 ## Summary
 - This PR implements an MVP integration for the Rokt Launcher script via mParticle’s initForwarder function. The changes ensure that the Rokt SDK is loaded  the resulting launcher is exposed globally for partner usage.

 ## Testing Plan
 - [x] Was this tested locally? If not, explain why.
 - This integration was validated locally using the npm run testEndToEnd command. During testing, I confirmed that the Rokt launcher script loads correctly and that the window.Rokt object is present. Additionally, the initForwarder function correctly assigns the launcher to the global scope, ensuring accessibility as expected.

 ## Reference Issue (For mParticle employees only.  Ignore if you are an outside contributor)
 - Closes https://go.mparticle.com/work/REPLACEME


.nvmrc
```diff
@@ -1 +1,5 @@
-v20.13.1
\ No newline at end of file
+<<<<<<< HEAD
+v20.13.1
+=======
+v20.13.1
+>>>>>>> development
```
Ethaywien, line 5: Seems a conflict diff got committed by accident.

src/initialization.js
```diff
@@ -1,42 +1,98 @@
+var roktLauncherScript = 'https://apps.rokt.com/wsdk/integrations/launcher.js';
+
 var initialization = {
-    name: 'insertSDKNameHere',
-/*  ****** Fill out initForwarder to load your SDK ******
+    name: 'RoktWsdk',
+    moduleId: 181,
```
crisryantan, line 5: Based from our discussion this morning @alexs-mparticle. Let me know if I need to change that though or if I should omit this for now as you also mentioned this might change

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,27 @@
+name: Dependabot Auto Merge
```
alexs-mparticle, line 1: I think you can remove this for now

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,32 @@
+name: Web Kit Build & Test
```
alexs-mparticle, line 1: You can remove this file

.github/workflows/release.yml
```diff
@@ -145,4 +145,4 @@ jobs:
             - name: Delete release branch
               if: ${{ github.event.inputs.dryRun == 'false' }}
               run: |
-                  git push --delete origin release/${{ github.run_number }}
\ No newline at end of file
+                  git push --delete origin release/${{ github.run_number }}
```
alexs-mparticle, line 148: Looks like this change is unnecessary. Is the linter complaining about it?

.travis.yml
```diff
@@ -0,0 +1,14 @@
+language: node_js
```
alexs-mparticle, line 1: Remove this file

README.md
```diff
@@ -1 +1,26 @@
-# mparticle-javascript-integration-rokt
\ No newline at end of file
+# mParticle JS Example Web Integration
```
alexs-mparticle, line 1: You should update this file to reflect any particular configuration settings from your SDK

dist/RoktWSDK-Kit.common.js
```diff
@@ -0,0 +1,829 @@
+'use strict';
```
alexs-mparticle, line 1: Please remove everything from `/dist` except for the `.keep` file. We'll build the bundle as part of the release.

alexs-mparticle (changes requested):

.github/workflows/release.yml
crisryantan: I'm not entirely sure. Tbh I think it's just my code editor default to leave a trailing whitespace at the end of every file. I can remove

README.md
crisryantan: For MVP purposes I will probably just leave it as it is since we are going to have a few more iterations until we have a clear set of requirements when we scale this

.github/workflows/release.yml
crisryantan: When running the `npm run lint:fix` command that was included in the boilerplate, it seems to add the trailing whitespace in js files. So I will keep those @alexs-mparticle

src/initialization.js
```diff
@@ -1,42 +1,98 @@
...
+                        window.Rokt &&
+                        typeof window.Rokt.createLauncher === 'function' &&
+                        window.Rokt.currentLauncher === undefined
+                    ) {
+                        window.Rokt.createLauncher({
+                            accountId: forwarderSettings.accountId,
+                            sandbox:
+                                forwarderSettings.sandbox !== undefined
+                                    ? forwarderSettings.sandbox
+                                    : true,
```
alexs-mparticle, line 51: Let's rename this to `sandboxMode` which will be how it is returned via the config. Also, the value will be `"True"` as a string, so you should update this check.
```suggestion
                            sandbox:
                                forwarderSettings.sandboxMode && forwarderSettings.sandboxMode  === 'True'
                                    ? true
                                    : false,
```

It might be cleaner if you just create variables to hold your settings at the top of this function, and just assign them directly here.

alexs-mparticle (changes requested):

src/initialization.js
```diff
@@ -1,42 +1,99 @@
...
+                        typeof window.Rokt.createLauncher === 'function' &&
+                        window.Rokt.currentLauncher === undefined
+                    ) {
+                        window.Rokt.createLauncher({
+                            accountId: forwarderSettings.accountId,
+                            sandbox:
+                                forwarderSettings.sandboxMode &&
+                                forwarderSettings.sandboxMode === 'True'
+                                    ? true
+                                    : false,
```
rmi22186, line 52: This could be simplified to 
```suggestion
                            sandbox: forwarderSettings.sandboxMode === 'True'
```
since if `forwarderSettings.sandboxMode` was `undefined` or `null`, this statement would return `false`

src/initialization.js
```diff
@@ -1,42 +1,99 @@
...
+                                if (window['Rokt'] && eventQueue.length > 0) {
+                                    for (
+                                        var i = 0;
+                                        i < eventQueue.length;
+                                        i++
+                                    ) {
+                                        processEvent(eventQueue[i]);
+                                    }
+                                    eventQueue = [];
+                                }
```
rmi22186, line 66: Is there a use case for an mparticle event to be passed to the Rokt SDK for processing? I don't think so.  It doesn't hurt to have this, but if Rokt never processes mParticle events, then this can be removed.
```suggestion
```

rmi22186 (commented):
great work getting this going!  i have some minor comments.  nothing blocking release though.

alexs-mparticle (approved):

PR #5, opened 2025-03-13
feat: add rokt preparitive iframe

crisryantan:
## Summary
We have recently conducted performance tests with mParticle and the different environments: [Performance Test Results](https://docs.google.com/spreadsheets/d/1oki65Ayuii_a0vySTLfSH5MDPfP5761S2cT0fPGwYsw/edit?gid=1556007761#gid=1556007761)
As you’ll see, there’s a bit more latency when loading Rokt through the mParticle kit. My understanding is that the page first has to load the mParticle Web SDK, which then injects the Rokt script. The extra steps in mParticle’s initialization plus loading Rokt Wsdk through kit leads to additional overhead. This explains why we’re seeing:

~50ms more in TTI compared to directly loading the Rokt WSDK
~200ms more in TTI when it’s Rokt WSDK “preloaded”

## Testing Plan
This is a proof of concept draft PR that I wanted to run by with @alexs-mparticle and @rmi22186 and see how we can improve performance when loading Rokt Wsdk through mparticle kit

### DIFF
--- a/dist/Rokt-Kit.common.js
+++ b/dist/Rokt-Kit.common.js
@@ -167,6 +167,26 @@ IdentityHandler.prototype.onSetUserIdentity = function (
 var identityHandler = IdentityHandler;
 
 var roktLauncherScript = 'https://apps.rokt.com/wsdk/integrations/launcher.js';
+var roktPreloadIframe = 'https://apps.rokt.com/wsdk/preload/index.html';
+
+function loadRoktPreloadIframe() {
+    if (document.getElementById('rokt-preload-iframe')) {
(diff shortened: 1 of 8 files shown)


dist/Rokt-Kit.common.js
```diff
@@ -194,6 +214,7 @@ var initialization = {
         _clientId
     ) {
         if (!testMode) {
+            loadRoktPreloadIframe();
```
crisryantan, line 217: Is there a point in time we can call this even earlier before the kit boots up?

PR #8, opened 2025-03-24
Feat: EXP-4851 - Add optimizely attribute fetch function

patrick-wu:
## Summary

Adding the following to the Rokt kit:
- ingesting new `settings` field that indicates whether there's a `onboardingExpProvider`
- new function for fetching optimizely experiments and transforming them to select-placement attributes
- before `selectPlacement()`, executes the Optimizely attribute fetch function and attach Optimizely attributes to the SP call

## Testing Plan
tested locally

### DIFF
--- a/src/Rokt-Kit.js
+++ b/src/Rokt-Kit.js
@@ -42,8 +42,7 @@ var constructor = function () {
     ) {
         var accountId = settings.accountId;
         var sandboxMode = settings.sandboxMode === 'True';
-        var enableOnboardingFlow = settings.enableOnboardingFlow === 'False';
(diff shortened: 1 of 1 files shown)


src/Rokt-Kit.js
```diff
@@ -74,6 +82,34 @@ var constructor = function () {
         }
     }
 
+    function fetchOptimizely() {
+        // return window.mParticle.ready(function () {
```
patrick-wu, line 86: i can't get this to work. From what I understand (which may be very incorrect), the `ready` function always return a void so it doesn't actually return anything in the inner anonymous function.

PR #9, opened 2025-03-24
feat: Add User Filtering to Kit

alexs-mparticle:
## Summary
Add User Filtering logic to Rokt Kit

## Testing Plan
- Using a sample app, attempt to pass in attributes to `window.mParticle.Rokt.selectPlacement` that have been filtered via the mParticle UI and verify the filtered attributes are not passed through.

### DIFF
--- a/src/Rokt-Kit.js
+++ b/src/Rokt-Kit.js
@@ -28,6 +28,13 @@ function attachLauncher(accountId, sandboxMode) {
             window.Rokt.currentLauncher = launcher;
             window.mParticle.Rokt.attachLauncher(launcher);
 
+            // Locally cache the launcher and filters
+            self.launcher = launcher;
+            self.filters = window.mParticle.Rokt.filters;
+
+            // Attaches the kit to the Rokt manager
+            window.mParticle.Rokt.attachKitself;
+
             self.isInitialized = true;
         })
         .catch(function (err) {
@@ -41,8 +48,22 @@ var constructor = function () {
     self.name = name;
     self.moduleId = moduleId;
     self.isInitialized = false;
+    self.launcher = null;
+    self.filters = {};
+    self.filteredUser = {};
+    self.userAttributes = {};
 
-    function initForwarder(settings, service, testMode) {
+    function initForwarder(
+        settings,
+        service,
+        testMode,
+        trackerId,
+        _userAttributes,
+        userIdentities,
+        appVersion,
+        appName,
+        customFlags
+    ) {
         var accountId = settings.accountId;
         var sandboxMode = window.mParticle.getEnvironment() === 'development';
 
@@ -50,6 +71,8 @@ var constructor = function () {
             attachLauncher(accountId, sandboxMode);
             return;
         }
+
+        self.userAttributes = _userAttributes;
 
         if (!window.Rokt || !(window.Rokt && window.Rokt.currentLauncher)) {
             var target = document.head || document.body;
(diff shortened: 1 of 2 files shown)


BrandonStalnaker (approved):
The logic looks good to me

src/Rokt-Kit.js
```diff
@@ -116,6 +181,19 @@ function isObject(val) {
     );
 }
 
+function mergeObjects() {
```
crisryantan, line 184: You could simplify this into two different options:

Option 1 (ES5-compatible)  
```
function mergeObjects() {
  var result = {};

  for (var i = 0; i < arguments.length; i++) {
    var obj = arguments[i];
    if (obj && typeof obj === 'object') {
      Object.assign(result, obj);
    }
  }

  return result;
}
```

Option 2 (Modern ES6+)  
Not sure how much modern api we could use with mParticle but if we could use ES6+ we could simplify this even further with something like:
```
function mergeObjects(...objects) {
  return Object.assign({}, ...objects);
}
```

alexs-mparticle: This is a good suggestion. However this mergeObject function is a common utility function we've included in all of our kits. Until we have transpiling set up for kits, we try to default to as "old school" as possible, particularly for utility functions.

src/Rokt-Kit.js
```diff
@@ -74,7 +98,48 @@ var constructor = function () {
         }
     }
 
+    function selectPlacements(options) {
+        var placementAttributes = mergeObjects(
+            options?.attributes,
```
alexs-mparticle, line 103: Note to self: refactor this to ES5

src/Rokt-Kit.js
```diff
@@ -74,7 +94,46 @@ var constructor = function () {
         }
     }
 
+    function selectPlacements(options) {
```
crisryantan, line 97: Do we want to add coverage on these?

src/Rokt-Kit.js
```diff
@@ -52,7 +65,14 @@ var constructor = function () {
                         .then(function (launcher) {
                             // Assign the launcher to a global variable for later access
                             window.Rokt.currentLauncher = launcher;
+
+                            // Locally cache the launcher and filters
+                            self.launcher = launcher;
+                            self.filters = window.mParticle.Rokt.filters;
+
+                            // Attaches the launcher and kit to the Rokt manager
                             window.mParticle.Rokt.attachLauncher(launcher);
```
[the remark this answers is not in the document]
rmi22186: seems like when you [attach the kit it then calls processMessageQueue](https://github.com/mParticle/mparticle-web-sdk/pull/1007/files#diff-7cb37e3efb8e0c8ba283a333407dd58bed345121a89f3b82014aabf2048cc61cR89) on the core SDK, so i think this is fine and we can remove attachLauncher completely.

src/Rokt-Kit.js
```diff
@@ -74,7 +94,46 @@ var constructor = function () {
         }
     }
 
+    function selectPlacements(options) {
+        var attributes = (options && options.attributes) || {};
```
rmi22186, line 98: let's add `identifier` to here as well https://docs.rokt.com/developers/integration-guides/web/library/select-placements-options.  It is optional but customers may already be passing this in.

https://docs.rokt.com/developers/integration-guides/web/library/select-placements-options

src/Rokt-Kit.js
```diff
@@ -74,7 +94,46 @@ var constructor = function () {
...
+        var placementAttributes = mergeObjects(attributes, self.userAttributes);
```
rmi22186, line 99: i would flip these, so that the attributes passed in from the customer override anything we have cached.

src/Rokt-Kit.js
rmi22186: actually i realize that identifier will just be passed through on line 109, right?

rmi22186: i would include a test where you are passing `identifier` through and confirm it is passed to `selectPlacements`

rmi22186 (changes requested):
looks good.  mostly nits.  tests?

src/Rokt-Kit.js
alexs-mparticle: Yes, I will be adding that in an upcoming PR

src/Rokt-Kit.js
alexs-mparticle: Yes, `identifier` is passed through. It should be defined in the interfaces for the Rokt Manager

rmi22186 (approved):

PR #11, opened 2025-03-25
fix: Add Test Framework

alexs-mparticle:
## Summary
- Introduces automated test framework to Kit

## Testing Plan
- Run automated tests

### DIFF
--- a/src/Rokt-Kit.js
+++ b/src/Rokt-Kit.js
@@ -18,14 +18,37 @@ var roktLauncherScript = 'https://apps.rokt.com/wsdk/integrations/launcher.js';
 var name = 'Rokt';
 var moduleId = 181;
 
+function attachLauncher(accountId, sandboxMode) {
+    window.Rokt.createLauncher({
+        accountId: accountId,
+        sandbox: window.mParticle.getEnvironment() === 'development',
+    })
+        .then(function (launcher) {
+            // Assign the launcher to a global variable for later access
+            window.Rokt.currentLauncher = launcher;
+            window.mParticle.Rokt.attachLauncher(launcher);
+
+            self.isInitialized = true;
+        })
+        .catch(function (err) {
+            console.error('Error creating Rokt launcher:', err);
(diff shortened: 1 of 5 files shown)


test/src/tests.js
```diff
@@ -83,203 +32,79 @@ describe('XYZ Forwarder', function () {
...
-        //
-        //     }
-        // };
-        //
-        // mParticle.forwarder.onUserIdentified(fakeUserStub);
-        //
-        // window.MockXYZForwarder.userId.should.equal('123');
-
-        done();
+const waitForCondition = function async(
```
rmi22186, line 92: i'd bring this higher up in the file.  feels weird to have it at the bottom.

rmi22186 (changes requested):
mostly lgtm.  tiny nits

src/Rokt-Kit.js
```diff
@@ -18,16 +18,39 @@ var roktLauncherScript = 'https://apps.rokt.com/wsdk/integrations/launcher.js';
 var name = 'Rokt';
 var moduleId = 181;
 
+function attachLauncher(accountId, sandboxMode) {
+    window.Rokt.createLauncher({
```
crisryantan, line 22: Minor, just wanted to check if we could start using async/await?

alexs-mparticle: We can't really use async/await because this file isn't being transpiled and we need kits to be as es5 as possible.

rmi22186 (approved):

src/Rokt-Kit.js
rmi22186: ^^ this.  technically even promises aren't ES5, but since your customer base is all.

As an FYI to Ryan, we've had to support ES5 primarily because some media customers use our web SDK on very old devices, like smart TVs from early 2000s.  Theoretically they'd probably not be using Rokt on those devices, but for now let's do it this way until we get kits properly transpiled.

PR #12, opened 2025-03-26
feat: replace sandboxMode check with mParticle.getEnvironment()

crisryantan:
## Summary
- Remove reference to sandbox when createLauncher is called in `src/Rokt-Kit.js`
- Update version to "0.0.3" in `package.json

## Testing Plan
- Haven't verified yet. Dependent on this [PR](https://github.com/mParticle/mparticle-web-sdk/pull/1009) being merged first

### DIFF
--- a/src/Rokt-Kit.js
+++ b/src/Rokt-Kit.js
@@ -46,7 +46,7 @@ var constructor = function () {
                 ) {
                     window.Rokt.createLauncher({
                         accountId: accountId,
-                        sandbox:  mParticle.getEnvironment() === 'development',
+                        sandbox: mParticle.getEnvironment() === 'development',
                     })
                         .then(function (launcher) {
                             // Assign the launcher to a global variable for later access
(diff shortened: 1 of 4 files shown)


crisryantan:
Disclaimer. Do not merge this until https://github.com/mParticle/mparticle-web-sdk/pull/1009 is merged and `@mparticle/web-sdk` is updated to the latest version

src/Rokt-Kit.js
```diff
@@ -47,7 +46,6 @@ var constructor = function () {
                 ) {
                     window.Rokt.createLauncher({
                         accountId: accountId,
-                        sandbox: sandboxMode,
```
alexs-mparticle, line 50: I think you can replace this line with `sandbox: mParticle.getEnvironment() === 'development',` and continue to use sandbox mode.

src/Rokt-Kit.js
```diff
@@ -47,7 +46,7 @@ var constructor = function () {
...
+                        sandbox: mParticle.getEnvironment() === 'development',
```
rmi22186, line 49: I would actually do `window.mParticle`.  Just in case the context ever changes for some reason (it shouldn't but this is just being safe and we do this in other kits.  Just like we call `window.Rokt` above) 
```suggestion
                        sandbox: window.mParticle.getEnvironment() === 'development',
```

PR #13, opened 2025-03-27
feat: Add User Attribute Mapping

alexs-mparticle:
## Summary
{provide a thorough description of the changes}

## Testing Plan


src/Rokt-Kit.js
```diff
@@ -116,6 +181,18 @@ function isObject(val) {
     );
 }
 
+function mergeObjects() {
```
alexs-mparticle, line 184: Note to self: Add tests around this function

PR #14, opened 2025-03-27
test: Add Test Coverage for selectPlacements call

alexs-mparticle:
## Summary
- Adds test coverage for selectPlacements call

## Testing Plan
- Run Automated Tests

### DIFF
--- a/test/src/tests.js
+++ b/test/src/tests.js
@@ -106,10 +106,6 @@ describe('Rokt Forwarder', () => {
             window.Rokt.accountId.should.equal('123456');
 
             window.Rokt.createLauncherCalled.should.equal(true);
-
-            await waitForCondition(
-                () => window.mParticle.Rokt.attachKitCalled === true
-            );
         });
     });
 
@@ -204,7 +200,6 @@ describe('Rokt Forwarder', () => {
 
         it('should process user attributes filters that are passed in via the config and selectPlacements options', async () => {
             window.mParticle.forwarder.filters = {
-                userAttributesFilters: [],
                 filterUserAttributes: function () {
                     return {
                         'user-attribute': 'user-attribute-value',
--- a/test/src/tests.js
+++ b/test/src/tests.js
@@ -198,7 +198,7 @@ describe('Rokt Forwarder', () => {
             });
         });
 
-        it('should process user attributes filters that are passed in via the config and selectPlacements options', async () => {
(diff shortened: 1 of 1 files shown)


test/src/tests.js
```diff
@@ -83,29 +85,163 @@ describe('Rokt Forwarder', () => {
...
+                true
+            );
+
+            window.Rokt.accountId.should.equal('123456');
+
+            window.Rokt.createLauncherCalled.should.equal(true);
+
+            await waitForCondition(
+                () => window.mParticle.Rokt.attachKitCalled === true
+            );
```
rmi22186, line 112: is this needed?  there's no assertion below this

test/src/tests.js
```diff
@@ -83,29 +85,163 @@ describe('Rokt Forwarder', () => {
...
+                attributes: {
+                    'user-attribute': 'user-attribute-value',
+                    'unfiltered-attribute': 'unfiltered-value',
+                },
+            });
+        });
+
+        it('should process user attributes filters that are passed in via the config and selectPlacements options', async () => {
+            window.mParticle.forwarder.filters = {
+                userAttributesFilters: [],
```
rmi22186, line 207: the test says it will process the user attribute filters that are passed on the config, but there are no filters here, so the attributes end up being the same as what is passed in.

rmi22186 (changes requested):

test/src/tests.js
alexs-mparticle: Technically the filters are used by `filterUserAttributes` via the Core SDK, and in this case we're mocking that function. The true purpose of this test is to verify that `selectPlacements` is correctly using the results of `filterUserAttributes`. Hope this change helps.

test/src/tests.js
```diff
@@ -83,29 +65,158 @@ describe('Rokt Forwarder', () => {
...
+                },
+            });
+        });
+
+        it('should filter user attributes through filterUserAttributes function before sending to selectPlacements', async () => {
+            window.mParticle.forwarder.filters = {
+                filterUserAttributes: function () {
+                    return {
+                        'user-attribute': 'user-attribute-value',
+                        'unfiltered-attribute': 'unfiltered-value',
```
rmi22186, line 186: change value of this to be diff from the one passed into selectPlacements

rmi22186 (approved):

PR #16, opened 2025-03-31
feat: add mparticle to launcher integration name

crisryantan:
## Summary
I've improved the version referencing in the mParticle-Rokt integration by:
1. Adding the `@rollup/plugin-replace` plugin to properly inject the package version at build time
2. Updated the rollup configuration to get actual version from package.json
3. Modified the string concatenation syntax in initialization.js

These changes ensure that the integration properly identifies itself with its version number when creating the Rokt launcher, allowing for better tracking and debugging of integration versions in production.

## Testing Plan
This has been tested by:
1. Verifying the syntax is valid JavaScript that works in all browser environments
2. Confirming the rollup configuration correctly imports and uses the package.json version


src/Rokt-Kit.js
```diff
@@ -22,6 +22,9 @@ function attachLauncher(accountId, sandboxMode) {
     window.Rokt.createLauncher({
         accountId: accountId,
         sandbox: sandboxMode,
+        integrationName: process.env.PACKAGE_VERSION
+            ? 'mParticle-' + process.env.PACKAGE_VERSION
```
mattbodle, line 26: There is another parameter for version: https://github.com/ROKT/sdk-web/blob/main/packages/web-sdk/integrations/src/launcher/models/private-integration-launcher-options.ts#L21

crisryantan: Closing the loop here, integrationVersion is specifically for launcher version after tracing through rokt wsdk code

src/Rokt-Kit.js
```diff
@@ -22,6 +22,9 @@ function attachLauncher(accountId, sandboxMode) {
     window.Rokt.createLauncher({
         accountId: accountId,
         sandbox: sandboxMode,
+        integrationName: process.env.PACKAGE_VERSION
```
alexs-mparticle, line 25: If I'm reading this right, the `integrationName` will just include the version of the Kit, correct?  I think there might be some value in collecting the version of the Core SDK as well, in case we notice some compatibility issues between the two.

You can grab that at runtime by collecting `window.mParticle.getVersion()`.

src/Rokt-Kit.js
```diff
@@ -22,6 +22,12 @@ function attachLauncher(accountId, sandboxMode) {
     window.Rokt.createLauncher({
         accountId: accountId,
         sandbox: sandboxMode,
+        integrationName:
```
crisryantan, line 25: @mattbodle just wanted to double-check with you guys, is WSDK team fine with this format?

mattbodle: is string? is good.

rmi22186 (approved):
LGTM pending confirmation of E2E testing now that's it's been rebased

PR #18, opened 2025-04-01
fix: Improve error handling and user attribute filtering in Rokt Kit

alexs-mparticle:
## Summary
- Adds some protection around filters and filtered users

## Testing Plan
- Use a sample app to verify interactions with filtered users

### DIFF
--- a/src/Rokt-Kit.js
+++ b/src/Rokt-Kit.js
@@ -62,11 +62,7 @@ var constructor = function () {
                     typeof window.Rokt.createLauncher === 'function' &&
                     window.Rokt.currentLauncher === undefined
                 ) {
-                    attachLauncher(accountId, sandboxMode).catch(function (
-                        err
-                    ) {
-                        console.error('Error attaching Rokt launcher:', err);
-                    });
+                    attachLauncher(accountId, sandboxMode);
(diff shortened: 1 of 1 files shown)


src/Rokt-Kit.js
```diff
@@ -138,8 +144,21 @@ var constructor = function () {
 
                 // Locally cache the launcher and filters
                 self.launcher = launcher;
-                self.filters = window.mParticle.Rokt.filters;
-                self.filteredUser = window.mParticle.Rokt.filters.filteredUser;
+
+                var roktFilters = window.mParticle.Rokt.filters;
+
+                if (!roktFilters) {
+                    console.warn('Rokt Kit: No filters have been set.');
```
rmi22186, line 151: should `roktFilters` be set to an empty object if there are no filters?  you'll have to update anywhere else `self.filters` is referenced to check for it first, like https://github.com/mparticle-integrations/mparticle-javascript-integration-rokt/blob/2bdb5e7803a05b070fa170a7a9807d785263de70/src/Rokt-Kit.js#L96C13-L100 and elsewhere.  do a global search for it

rmi22186 (approved):

PR #20, opened 2025-04-07
feat: Add optimizely attribute fetching functionality

patrick-wu:
Add Optimizely attribute-fetching function to Kit, invoke right before selectPlacement() calls

## Summary
{provide a thorough description of the changes}

## Testing Plan

### DIFF
--- a/dist/Rokt-Kit.common.js
+++ b/dist/Rokt-Kit.common.js
@@ -210,10 +210,13 @@ var constructor = function () {
             });
 
         try {
-            if (forwarders.length > 0 || window.optimizely) {
+            if (forwarders.length > 0 && window.optimizely) {
                 // Get the state object
                 var optimizelyState = window.optimizely.get('state');
-                if (!optimizelyState || !optimizelyState.getActiveExperimentIds) {
+                if (
+                    !optimizelyState ||
+                    !optimizelyState.getActiveExperimentIds
+                ) {
                     return {};
                 }
                 // Get active experiment IDs
--- a/dist/Rokt-Kit.common.js
+++ b/dist/Rokt-Kit.common.js
@@ -28,7 +28,6 @@ var constructor = function () {
     self.name = name;
     self.moduleId = moduleId;
     self.isInitialized = false;
-
     self.launcher = null;
     self.filters = {};
     self.filteredUser = {};
@@ -44,7 +43,6 @@ var constructor = function () {
         var accountId = settings.accountId;
         var sandboxMode = window.mParticle.getEnvironment() === 'development';
         self.userAttributes = filteredUserAttributes;
-        self.onboardingExpProvider = settings.onboardingExpProvider;
 
         if (testMode) {
             attachLauncher(accountId, sandboxMode);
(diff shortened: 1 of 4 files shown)


src/Rokt-Kit.js
```diff
@@ -187,6 +198,40 @@ var constructor = function () {
...
     // mParticle Kit Callback Methods
+    function fetchOptimizely() {
+        var forwarders = window.mParticle
+            ._getActiveForwarders()
+            .filter(function (forwarder) {
+                return forwarder.name === 'Optimizely';
+            });
+
+        try {
+            if (forwarders.length > 0 || window.optimizely) {
```
alexs-mparticle, line 209: I think logical AND would be more appropriate, right?

```suggestion
            if (forwarders.length > 0 && window.optimizely) {
```

src/Rokt-Kit.js
```diff
@@ -187,6 +198,40 @@ var constructor = function () {
...
+                // Get the state object
+                var optimizelyState = window.optimizely.get('state');
```
alexs-mparticle, line 211: I think you should add some handlers in case `optimizelyState` returns `undefined`.

src/Rokt-Kit.js
```diff
@@ -187,6 +198,40 @@ var constructor = function () {
     this.selectPlacements = selectPlacements;
 
     // mParticle Kit Callback Methods
+    function fetchOptimizely() {
```
alexs-mparticle, line 201: Now that we have a testing framework, I would like to see some tests around this logic. I can help you with mocking if you need it.

alexs-mparticle (changes requested):
Great work @patrick-wu! Appreciate you rebasing and resubmitting. So far I only have some minor nits but his looks good.

src/Rokt-Kit.js
patrick-wu: but what about the case where the optimizely sdk wasn't loaded via the kit setup, since the client can still load it directly?

[mparticle doc on optimizely integration](https://docs.mparticle.com/integrations/optimizely/event/#web)
> While allowing mParticle to automatic load Optimizely reduces the amount of code you need to write, you may choose to initialize Optimizely yourself to prevent “page flashing” in the case where an Optimizely experiment is expected to alter the UI immediately on load. You can [read more about this concern here](https://help.optimizely.com/Set_Up_Optimizely/Implement_the_one-line_snippet_for_Optimizely_X) and make the choice that’s best for your setup.

i guess even if they load it themselves, we'd still have a forwarder?

rmi22186: forwarder would still exist yes.

patrick-wu: ok got it. in that case yeah, i'll modify this to &&

src/Rokt-Kit.js
patrick-wu: @alexs-mparticle can you take a look if these test cases are sufficient?

dist/Rokt-Kit.common.js
```diff
@@ -120,9 +122,18 @@ var constructor = function () {
 
         self.userAttributes = filteredAttributes;
 
-        var selectPlacementsAttributes = mergeObjects(filteredAttributes, {
-            mpid: mpid,
-        });
+        var optimizelyAttributes =
```
alexs-mparticle, line 125: Please remove the dist files from the commit. These are built by the CI/CD process.

test/src/tests.js
```diff
@@ -413,4 +416,205 @@ describe('Rokt Forwarder', () => {
...
+                    'rokt.custom.optimizely'
+                );
+            });
+
+            it('should return empty object when Optimizely state is missing required methods', async () => {
+                setupInvalidOptimizelyMock({
+                    getVariationMap: function () {
+                        return {};
+                    },
+                    // Missing getActiveExperimentIds
```
alexs-mparticle, line 591: Nitpick: Updating comment for clarity
```suggestion
                    // Mocking a scenario for when getActiveExperimentIds() method is missing
```

alexs-mparticle (changes requested):

PR #21, opened 2025-04-10
feat: POC for supporting Optimizely fullstack / feature experimentation allocations

patrick-wu:
## Summary
Adding support for Optimizely Full Stack / Feature Experimentation as 3rd party experimentation platforms we can integrate with.

This is more complex than Optimizely Web because unlike OW, the Full Stack / Feature Exp SDK does not keep the state of the current session's allocation.

Instead, it can guarantee to return consistent allocation, *GIVEN* an 'ID' and any additional optional attribute for consistent hashing / allocation, which mean **the complexity now is we'll need to pass in the allocation ID and attributes into the `selectPlacement` call somehow**.

## Solution Presented in POC

We pass these things in with specific keys that we look for, use them to generate allcoations, and remove them from the `filteredAttributes` object before passing it to the `selectPlacement` call.

## Testing Plan

### DIFF
--- a/src/Rokt-Kit.js
+++ b/src/Rokt-Kit.js
@@ -18,6 +18,12 @@ var roktLauncherScript = 'https://apps.rokt.com/wsdk/integrations/launcher.js';
 var name = 'Rokt';
 var moduleId = 181;
 
+var OnboardingExpProvider = {
+    OPTIMIZELY_FEATURE_EXP: 'OptimizelyFeatureExp',
+    OPTIMIZELY_WEB: 'OptimizelyWeb',
+    NONE: 'None',
+};
+
 var constructor = function () {
     var self = this;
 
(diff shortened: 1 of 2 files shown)


src/Rokt-Kit.js
```diff
@@ -118,14 +124,43 @@ var constructor = function () {
...
+                        return key.indexOf('optimizely.') !== 0;
+                    })
+                    .reduce(function (obj, key) {
+                        obj[key] = filteredAttributes[key];
+                        return obj;
+                    }, {});
+                experimentAttributes = fetchOptimizelyFeatureExpAllocations(
+                    userId,
+                    attributes
+                );
```
patrick-wu, line 155: basically, for Optimizely Full Stack / Feature Experimentation, you need to pass in the user information (an ID, plus whatever additional attributes) to the SDK when trying to retrieve the bucketing information of the current user.

This kind of strategy is stateless (i.e. it doesn't keep track of what the allocation is) but idempotent, so we have to somehow pass in the user ID + attributes used when deciding

src/Rokt-Kit.js
```diff
@@ -238,6 +273,39 @@ var constructor = function () {
...
+            .filter(function (forwarder) {
+                return forwarder.name === 'Optimizely';
+            });
+
+        if (forwarders.length > 0 && window.optimizelyClient) {
+            var user = window.optimizelyClient.createUserContext(
+                userId,
+                attributes
+            );
+            var config = window.optimizelyClient.getOptimizelyConfig();
```
rmi22186, line 289: same comment regarding this one. check for `window.optimizelyClient.hasOwnProperty('getOptimizelyConfig')`

src/Rokt-Kit.js
```diff
@@ -305,4 +373,5 @@ if (window && window.mParticle && window.mParticle.addForwarder) {
 
 module.exports = {
     register: register,
+    OnboardingExpProvider: OnboardingExpProvider,
```
rmi22186, line 376: is this necessary?  If it's only necessary for the tests, I think you can just duplicate the values in the test itself.

This is generally just used for an end developer to use.

rmi22186 (changes requested):
couple of small suggestions.  Also can you add a test for when `OPTIMIZELY_FEATURE_EXP`.  There's a lot of complicated logic in here, so it's hard to know if it works or not and it's unclear what optimizely will return. You'll have to mock `mParticle._getActiveForwarders`, as well as a few optimizely methods that you are using.   let me know if you need help with this and we can book like 15-30 min.

patrick-wu:
closing this one as we've aligned on the other approach with the Rokt Manager

PR #25, opened 2025-04-23
feat: create example app to live preview changes in kit and to use for running e2e tests

crisryantan:
## Summary
- This change is for us to create an example app to live preview changes in the kit and to use for running end-to-end tests.

## Testing Plan
- Tested example app on local machine
- Added readme guide to run the example app locally

Screenshot 2025-04-29 at 3 57 02 pm


example-app/src/index.js
```diff
@@ -0,0 +1,165 @@
...
+        const apiKey = process.env.ROKT_API_KEY;
+
+        // Add Rokt kit configuration
+        mParticleConfig.kitConfigs.push({
+            name: 'Rokt',
+            moduleId: 181,
+            isDebug: true,
+            isVisible: true,
+            settings: {
+                accountId: process.env.ROKT_ACCOUNT_ID
```
crisryantan, line 137: For some reason, this doesn't work, @rmi22186 and @alexs-mparticle, just wanted to check-in if im doing anything wrong here

rmi22186: maybe you discussed this with alex already - but the reason it doesn't work is because you are relying on the snippet to load the kit configuration and what you are doing is more of an advanced use case.  if you self host, you can provide your own kit configuration like you are here, but this is ilkely a very low frequency use tool for a partner who does not want the UI to dynamically update their configuration.  just some background, we provided this option for a financial services company who wanted a strict lock down for everything on their site.  

if you want to pass the kitConfigs, you would need to self host, AND set `mParticleConfig.requestConfig = false`, then it will use the config that you pass in.  there isn't a way to pass a kit config in if you are using the snippet.

however, since partners should be using the snippet, the README should really just say that this app will only work if they have a properly configured Rokt configuration (which their customer team will do for them)

rmi22186: wait, i did just see that you have the `requestConfig` as being `false`.  sorry, i was confused as to if this was supposed to be a snippet or self hosted example since the readme says `Snippet Example`.

jklafka-rokt (approved):

PR #26, opened 2025-04-23
feat: Add Extensions to Rokt Kit

alexs-mparticle:
## Summary
- Adds support for vNext Extensions via the Rokt Kit
- Requires https://github.com/mParticle/mparticle-web-sdk/pull/1025 to be merged

## Testing Plan
- Using a sample app with placeholder divs, for example an `offer-container`, `cancel-button` and `confirm-button`.
- Load the SDK and Kit with extensions via a kit config
- Fire a select placement call
- Execute a call to `setExtensionData` using the following in the dev console:
```
window.mParticle.Rokt.setExtensionData({
  '<extension-name>': {
    confirmBtn: '#confirm-button',
    cancelBtn: '#cancel-button',
    container: '#offer-container'
  }
});
```
- Verify using the network tab that calls are made to `https://apps.rokt.com/v1/events` with:
  - `SignalPartnerImpression `
  - `SignalPartnerResponse`
  - `SignalPartnerDismissal`
- ...are made upon the initial impression, confirm button click and cancel button click, respectively.

### DIFF
--- a/src/Rokt-Kit.js
+++ b/src/Rokt-Kit.js
@@ -199,7 +199,13 @@ var constructor = function () {
         return self.launcher.selectPlacements(selectPlacementsOptions);
     }
 
-    // TODO: Add JSDocs
+    /**
+     * Sets extension data for Rokt Web SDK
+     * @param {Object} partnerExtensionData - The extension data object containing:
+     * - [extensionName] {string}: Name of the extension
+     * - [extensionName].options {Object}: Key-value pairs of options for the extension
+     * @returns {void} Nothing is returned
+     */
     function setExtensionData(partnerExtensionData) {
         if (!isInitialized()) {
             console.error('Rokt Kit: Not initialized');
@@ -254,9 +260,6 @@ var constructor = function () {
                 }
                 // Attaches the kit to the Rokt manager
                 window.mParticle.Rokt.attachKit(self);
-
-                // TODO: Add to Core SDK
-                window.mParticle.Rokt.setExtensionData = self.setExtensionData;
 
                 self.isInitialized = true;
             })
--- a/src/Rokt-Kit.js
+++ b/src/Rokt-Kit.js
@@ -265,7 +265,6 @@ var constructor = function () {
                 window.mParticle.Rokt.attachKit(self);
 
                 self.isInitialized = true;
-                console.warn('Rokt TRACE: launcher called');
             })
             .catch(function (err) {
                 console.error('Error creating Rokt launcher:', err);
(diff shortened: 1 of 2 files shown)


src/Rokt-Kit.js
```diff
@@ -29,14 +35,21 @@ var constructor = function () {
...
     self.filteredUser = {};
     self.userAttributes = {};
+    self.testHelpers = null;
 
     /**
-     * Passes attributes to the Rokt Web SDK for client-side hashing
-     * @see https://docs.rokt.com/developers/integration-guides/web/library/integration-launcher#hash-attributes
-     * @param {Object} attributes - The attributes to be hashed
-     * @returns {Promise<Object|null>} A Promise resolving to the
-     * hashed attributes from the launcher, or `null` if the kit is not initialized
```
rmi22186, line 38: looks like this got deleted.  bring it back for `hashAttributes`

src/Rokt-Kit.js
```diff
@@ -55,8 +68,15 @@ var constructor = function () {
         var accountId = settings.accountId;
         self.userAttributes = filteredUserAttributes;
         self.onboardingExpProvider = settings.onboardingExpProvider;
+        self.vNextExtensions = extractvNextExtensions(settings.vNextExtensions);
```
rmi22186, line 71: if this is only being used once, there isn't a reason for it to be on the class.  it can just be declared as a variable nd passed into line 73, right?

src/Rokt-Kit.js
```diff
@@ -155,6 +175,25 @@ var constructor = function () {
...
+     */
+    function setExtensionData(partnerExtensionData) {
+        if (!isInitialized()) {
+            console.error('Rokt Kit: Not initialized');
+            return;
+        }
+
+        // TODO: Should we check if select placements has been called?
+        // Some extensions seem to need that to happen first
+        // TODO: Should we attach the Rokt SDK to the kit as well?
```
rmi22186, line 193: not if there isn't a reason to.  the paradigm is currently core calls kit. kit calls SDK.  I think we should keep it that way.

src/Rokt-Kit.js
```diff
@@ -243,6 +245,7 @@ var constructor = function () {
                 window.mParticle.Rokt.attachKit(self);
 
                 self.isInitialized = true;
+                console.warn('Rokt TRACE: launcher called');
```
rmi22186, line 248: nit: launcher isn't relaly being called here, it's more like Rokt is now ready, right?

alexs-mparticle: You can ignore that. It was a debug log I forgot to remove.

rmi22186 (approved):
approving, but we need to confirm with ignacio/sarthi how long adding a new cusom json mapping would take still

jklafka-rokt (approved):

rmi22186 (approved):

PR #35, opened 2025-06-16
feat: Support adding identities to select placement attributes

rmi22186:
## Summary
We want to add all user identities to the attributes that get passed to `selectPlacements`

## Testing Plan
Added unit tests.  Tested in local app also.

### DIFF
--- a/test/src/tests.js
+++ b/test/src/tests.js
@@ -597,7 +597,8 @@ describe('Rokt Forwarder', () => {
                     {
                         // These should be filtered out
                         'blocked-attribute': 'blocked-value',
-                        'initial-user-attribute': 'initial-user-attribute-value',
+                        'initial-user-attribute':
+                            'initial-user-attribute-value',
 
                         // This should be updated
                         'changed-attribute': 'old-value',


src/Rokt-Kit.js
```diff
@@ -177,6 +209,7 @@ var constructor = function () {
 
     function onUserIdentified(filteredUser) {
         self.filteredUser = filteredUser;
+        self.filters.filteredUser = filteredUser;
```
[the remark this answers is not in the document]
rmi22186: I removed self.filteredUser since it isn't used anywhere.  per our zoom call, RoktManager sets `RoktManager.filters.filteredUser`, so I am just updating that

src/Rokt-Kit.js
```diff
@@ -214,11 +227,8 @@ var constructor = function () {
                         console.warn(
                             'Rokt Kit: No filtered user has been set.'
                         );
-                    } else {
-                        self.filteredUser = roktFilters.filteredUser;
                     }
```
rmi22186, line 230: removing this line to avoid having 2 filtered users, 1 on `self.filteredUser`, and 1 on `self.filters.filteredUser` which is already set in line 212

test/src/tests.js
```diff
@@ -620,6 +623,309 @@ describe('Rokt Forwarder', () => {
...
+
+                window.Rokt.selectPlacementsOptions.attributes.should.deepEqual(
+                    {
+                        'test-attribute': 'test-value',
+                        mpid: 'abc',
+                    }
+                );
+            });
+
+            it('should handle case when userAttributes is null but userIdentities exist', async () => {
```
alexs-mparticle, line 702:
```suggestion
            it('should send userIdentities when userAttributes is null but userIdentities exists', async () => {
```

test/src/tests.js
```diff
@@ -620,6 +623,309 @@ describe('Rokt Forwarder', () => {
...
+                };
+                window.mParticle.forwarder.launcher = {
+                    selectPlacements: function (options) {
+                        window.mParticle.Rokt.selectPlacementsOptions = options;
+                        window.mParticle.Rokt.selectPlacementsCalled = true;
+                    },
+                };
+            });
+
+            it('should handle case when userIdentities is null but userAttributes exist', async () => {
```
alexs-mparticle, line 645:
```suggestion
            it('should send userAttributes if userIdentities is null but userAttributes exists', async () => {
```

alexs-mparticle (changes requested):

PR #37, opened 2025-07-14
fix: Handle invalid Extension Settings

alexs-mparticle:
## Summary

- Handle cases where setting string may come back incorrectly, such as `NONE` or `NULL`.

## Testing Plan
- Using a sample app with local overrides, test different variations of invalid `roktExtension` settings.

### DIFF
--- a/src/Rokt-Kit.js
+++ b/src/Rokt-Kit.js
@@ -404,7 +404,7 @@ function parseSettingsString(settingsString) {
 }
 
 function extractRoktExtensions(settingsString) {
-    var settings = parseSettingsString(settingsString);
+    var settings = settingsString ? parseSettingsString(settingsString) : [];
     var roktExtensions = [];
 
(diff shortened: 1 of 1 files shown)


src/Rokt-Kit.js
```diff
@@ -398,14 +398,15 @@ function parseSettingsString(settingsString) {
...
-        throw new Error('Settings string contains invalid JSON');
+        console.error('Settings string contains invalid JSON');
     }
+    return [];
 }
 
 function extractRoktExtensions(settingsString) {
-    var settings = settingsString ? parseSettingsString(settingsString) : [];
-
+    var settings = parseSettingsString(settingsString);
```
rmi22186, line 407: I think this should still have the conditional.  otherwise everyone immediately will start to see a console.error about there not being invalid json right?, right?

rmi22186 (approved):

PR #40, opened 2025-08-13
Feat/sqdsdks 7556 hashed email for rokt

rmi22186:
## Summary
This PR adjusts behavior of how we handle `other`. Previously, `other` from selectAttributes was dropped as it was assumed to be the customer passing int a hashed email.  However, product decided we should treat `other` as any other attribute.  

This PR also removes `email` in addition to `other` as an identity if `other` is passed in.
Finally, if selectPlacement passes in `emailsha256`, we remove the `email` identity.

## Testing Plan
Unit tests added.

### DIFF
--- a/src/Rokt-Kit.js
+++ b/src/Rokt-Kit.js
@@ -155,14 +155,14 @@ var constructor = function () {
         return data;
     }
 
-    // function replaceEmailIdentityWithEmailsha256(_data) {
-    //     var data = mergeObjects({}, _data || {});
-    //     if (_data.hasOwnProperty(EMAIL_SHA256_IDENTITY)) {
-    //         delete data[EMAIL_IDENTITY];
-    //     }
+    function sanitizeIdentities(_data) {
+        var data = mergeObjects({}, _data || {});
+        if (_data.hasOwnProperty(EMAIL_SHA256_IDENTITY)) {
+            delete data[EMAIL_IDENTITY];
+        }
 
-    //     return data;
-    // }
+        return data;
+    }
 
     /**
      * Selects placements for Rokt Web SDK with merged attributes, filters, and experimentation options
@@ -209,7 +209,6 @@ var constructor = function () {
                 : {};
 
         var filteredUserIdentities = returnUserIdentities(filteredUser);
-        console.log(filteredUserIdentities);
 
         var selectPlacementsAttributes = mergeObjects(
             filteredUserIdentities,
@@ -219,9 +218,9 @@ var constructor = function () {
                 mpid: mpid,
             }
         );
-        console.log(selectPlacementsAttributes);
+
         var selectPlacementsOptions = mergeObjects(options, {
-            attributes: selectPlacementsAttributes,
+            attributes: sanitizeIdentities(selectPlacementsAttributes),
         });
 
         return self.launcher.selectPlacements(selectPlacementsOptions);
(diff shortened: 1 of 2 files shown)


src/Rokt-Kit.js
```diff
@@ -140,14 +141,24 @@ var constructor = function () {
...
-        return replaceOtherWithEmailsha256(userIdentities);
+        return replaceOtherIdentityWithEmailsha256(userIdentities);
     }
 
-    function replaceOtherWithEmailsha256(_data) {
+    function replaceOtherIdentityWithEmailsha256(_data) {
         var data = mergeObjects({}, _data || {});
         if (_data.hasOwnProperty(OTHER_IDENTITY)) {
             data[EMAIL_SHA256_IDENTITY] = _data[OTHER_IDENTITY];
             delete data[OTHER_IDENTITY];
```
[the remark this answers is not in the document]
rmi22186: correct. I decided to make 2 distinct functions to make it clear.  good question though

src/Rokt-Kit.js
```diff
@@ -153,6 +154,15 @@ var constructor = function () {
         return data;
     }
 
+    function sanitizeIdentities(_data) {
```
alexs-mparticle, line 157: I think `sanitize` is too vague and generic.

```suggestion
    function sanitizeEmailIdentities(_data) {
```

alexs-mparticle (changes requested):

src/Rokt-Kit.js
```diff
@@ -85,6 +88,13 @@ var constructor = function () {
             placementEventMapping
         );
 
+        // Set dynamic OTHER_IDENTITY based on server settings
+        // Convert to lowercase since server sends TitleCase (e.g., 'Other' -> 'other')
+        if (settings.hashedEmailUserIdentityType) {
+            MAPPED_EMAIL_SHA256_IDENTITY =
```
alexs-mparticle, line 94: I would make `MAPPED_EMAIL_SHA256_IDENTITY` a standard variable if you plan on assigning it to a value. This looks like a constant as is, so it shouldn't be assigned a value like this.

alexs-mparticle (changes requested):

src/Rokt-Kit.js
```diff
@@ -151,8 +161,7 @@ var constructor = function () {
         }
 
         var userIdentities = filteredUser.getUserIdentities().userIdentities;
-
-        return replaceOtherWithEmailsha256(userIdentities);
+        return replaceOtherIdentityWithEmailsha256(userIdentities);
```
alexs-mparticle, line 164:
```suggestion

        return replaceOtherIdentityWithEmailsha256(userIdentities);
```

test/src/tests.js
```diff
@@ -1374,20 +1375,211 @@ describe('Rokt Forwarder', () => {
...
+                window.Rokt.selectPlacementsOptions.attributes.should.deepEqual(
+                    {
+                        'test-attribute': 'test-value',
+                        emailsha256: 'hashed-customer-id-value', // mapped from customerid in userIdentities
+                        mpid: '789',
+                    }
+                );
+            });
+
+            it('should not set emailsha256 on final select placements attributes when hashedEmailUserIdentityType is Unassigned', async () => {
```
alexs-mparticle, line 1458:
```suggestion
            it('should NOT set emailsha256 on final select placements attributes when hashedEmailUserIdentityType is Unassigned', async () => {
```

alexs-mparticle (changes requested):

alexs-mparticle (approved):

PR #41, opened 2025-08-28
feature: Map events to an attribute flag for selectPlacements

alexs-mparticle:
## Summary

Map events to an attribute flag for selectPlacements

## Testing Plan
- Use a sample app and map custom events to an attribute flag using the mParticle UI
- Fire the mapped event using the sample app
- Verify that the selectPlacement call is made with the attribute flag
- Restart your browser and verify that values persist
- End your session and verify that the values have cleared

### DIFF
--- a/src/Rokt-Kit.js
+++ b/src/Rokt-Kit.js
@@ -29,6 +29,8 @@ var constructor = function () {
     self.filters = {};
     self.userAttributes = {};
     self.testHelpers = null;
+    self.placementEventMappingLookup = {};
+    self.store = null;
 
     /**
      * Generates the Rokt launcher script URL with optional domain override and extensions
@@ -74,8 +76,20 @@ var constructor = function () {
     ) {
         var accountId = settings.accountId;
         var roktExtensions = extractRoktExtensions(settings.roktExtensions);
-        self.userAttributes = filteredUserAttributes;
+        self.userAttributes = filteredUserAttributes || {};
         self.onboardingExpProvider = settings.onboardingExpProvider;
+
+        var placementEventMapping = parseSettingsString(
+            settings.placementEventMapping
+        );
+        self.placementEventMappingLookup = generateMappedEventLookup(
+            placementEventMapping
+        );
+
+        if (window.mParticle.Rokt.store) {
+            self.store = window.mParticle.Rokt.store;
+        }
+
         var domain = window.mParticle.Rokt.domain;
         var launcherOptions = mergeObjects(
             {},
@@ -89,6 +103,9 @@ var constructor = function () {
             self.testHelpers = {
                 generateLauncherScript: generateLauncherScript,
                 extractRoktExtensions: extractRoktExtensions,
+                hashEventMessage: hashEventMessage,
+                parseSettingsString: parseSettingsString,
+                generateMappedEventLookup: generateMappedEventLookup,
             };
             attachLauncher(accountId, launcherOptions);
             return;
@@ -199,10 +216,15 @@ var constructor = function () {
 
         var filteredUserIdentities = returnUserIdentities(filteredUser);
 
+        var localSessionAttributes = self.store
+            ? self.store.getLocalSessionAttributes()
+            : {};
+
         var selectPlacementsAttributes = mergeObjects(
             filteredUserIdentities,
             replaceOtherWithEmailsha256(filteredAttributes),
             optimizelyAttributes,
+            localSessionAttributes,
             {
                 mpid: mpid,
             }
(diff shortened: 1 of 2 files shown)


src/Rokt-Kit.js
```diff
@@ -231,6 +253,27 @@ var constructor = function () {
...
+            event.EventCategory,
+            event.EventName
+        );
+
+        if (self.placementEventMappingLookup[hashedEvent]) {
+            var mappedValue = self.placementEventMappingLookup[hashedEvent];
+            if (self.store) {
+                var attrs = {};
+                attrs[mappedValue] = true;
+                self.store.setLocalSessionAttributes(attrs);
```
rmi22186, line 272:
```suggestion
                window.mParticle.Rokt.setLocalSessionAttribute(mappedValue, true);
```

test/src/tests.js
```diff
@@ -636,6 +709,53 @@ describe('Rokt Forwarder', () => {
...
+                                map: '-5208850776883573773',
+                                maptype: 'EventClass.Id',
+                                value: 'foo-mapped-flag',
+                            },
+                            {
+                                jsmap: '1838502119',
+                                map: '1324617889422969328',
+                                maptype: 'EventClass.Id',
+                                value: 'ad_viewed_test',
+                            },
```
rmi22186, line 734: Is having an actual mapping necessary here if we already have items in `localSessionAttributes`.  And if it's for illustrative purposes, then i think you should have the `value`s in the mapping match the keys for what is in `localSessionAttributes`

test/src/tests.js
```diff
@@ -1593,4 +1713,214 @@ describe('Rokt Forwarder', () => {
...
+                    {
+                        jsmap: null,
+                        map: 'last_name',
+                        maptype: 'UserAttributeClass.Name',
+                        value: 'lastname',
+                    },
+                ]);
+        });
+
+        it('should parse a settings string with a number value', () => {
```
rmi22186, line 1865: Is the name of this test misleading, or are all the values in the assertion all strings, and none are numbers, which is what I'm assuming is the point of this test?  

although it would be a stringified number and not an actual number since the server only returns strings for the mappings

test/src/tests.js
```diff
@@ -1593,4 +1713,214 @@ describe('Rokt Forwarder', () => {
...
+
+        it('returns an empty array if the settings string is empty', () => {
+            const settingsString = '';
+
+            window.mParticle.forwarder.testHelpers
+                .parseSettingsString(settingsString)
+                .should.deepEqual([]);
+        });
+
+        it('throws an error message if the settings string is not a valid JSON', () => {
```
rmi22186, line 1901: the test does not match the name of this test

src/Rokt-Kit.js
```diff
@@ -199,10 +211,14 @@ var constructor = function () {
 
         var filteredUserIdentities = returnUserIdentities(filteredUser);
 
+        var localSessionAttributes =
+            window.mParticle.Rokt.getLocalSessionAttributes() || {};
```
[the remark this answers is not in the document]
rmi22186: @alexs-mparticle I concur with the robot here

test/src/tests.js
```diff
@@ -1593,4 +1718,222 @@ describe('Rokt Forwarder', () => {
...
+                },
+                filteredUser: {
+                    getMPID: function () {
+                        return '123';
+                    },
+                },
+            };
+        });
+
+        it('set a session selection attribute if the event is a mapped placement event', async () => {
```
rmi22186, line 1813:
```suggestion
        it('set a local session attribute if the event is a mapped placement event', async () => {
```

test/src/tests.js
```diff
@@ -1593,4 +1718,222 @@ describe('Rokt Forwarder', () => {
...
+                    {
+                        jsmap: null,
+                        map: 'last_name',
+                        maptype: 'UserAttributeClass.Name',
+                        value: 'lastname',
+                    },
+                ]);
+        });
+
+        it('should parse a settings string with a stringified number value correctly', () => {
```
rmi22186, line 1878: I'm still unclear about the meaning of this test.  what about a stringified number is being tested?  saw that you resolved the original comment but didn't see any changes

alexs-mparticle: The purpose of this test is to verify that if a setting string is parsed, the string hashes in the `jsmap` and `map` properties are not converted to numbers, but remain as strings.

For example, `&quot;jsmap&quot;:&quot;-1484452948&quot;` should be come `jsmap: '-1484452948'` (sting) and not `-1484452948` (number)

test/src/tests.js
```diff
@@ -1593,4 +1718,222 @@ describe('Rokt Forwarder', () => {
...
```
rmi22186, line 1878:
```suggestion
        it('should convert a jmap number value to a stringified number when parsed', () => {
```

PR #43, opened 2025-09-06
fix: Handle cases where local session storage is not available

alexs-mparticle:
## Summary
- Handles cases where local session storage is not available

## Testing Plan
- Using a sample app, remove `getLocalSessionAttributes` from the mParticle.Rokt namespace to verify an error is not thrown.

### DIFF
--- a/src/Rokt-Kit.js
+++ b/src/Rokt-Kit.js
@@ -155,6 +155,17 @@ var constructor = function () {
         return replaceOtherWithEmailsha256(userIdentities);
     }
 
+    function returnLocalSessionAttributes() {
+        if (
+            !window.mParticle.Rokt ||
+            typeof window.mParticle.Rokt.getLocalSessionAttributes !==
+                'function'
+        ) {
+            return {};
+        }
+        return window.mParticle.Rokt.getLocalSessionAttributes();
+    }
+
     function replaceOtherWithEmailsha256(_data) {
         var data = mergeObjects({}, _data || {});
         if (_data.hasOwnProperty(OTHER_IDENTITY)) {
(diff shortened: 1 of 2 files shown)


src/Rokt-Kit.js
```diff
@@ -494,6 +499,16 @@ function hashEventMessage(messageType, eventType, eventName) {
...
+function isEmpty(value) {
+    return (
+        value === null ||
+        value === undefined ||
+        (typeof value === 'object' && Object.keys(value).length === 0) ||
+        (typeof value === 'string' && value.length === 0) ||
+        (Array.isArray(value) && value.length === 0)
+    );
+}
+
```
rmi22186, line 511: This is _probably_ okay, but I do see an `isEmpty` helper in our other kits like [Braze](https://github.com/mparticle-integrations/mparticle-javascript-integration-braze/blob/master-v5/src/BrazeKit-dev.js#L1175-L1177) which matches our TS version in [core](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/utils.ts#L266-L267)

src/Rokt-Kit.js
```diff
@@ -155,6 +155,17 @@ var constructor = function () {
         return replaceOtherWithEmailsha256(userIdentities);
     }
 
+    function returnLocalSessionAttributes() {
+        if (
+            !window.mParticle.Rokt ||
+            typeof window.mParticle.Rokt.getLocalSessionAttributes !==
+                'function'
+        ) {
```
rmi22186, line 163: We should also add
```
isEmpty(self.placementEventMappingLookup
```

here since there isn't a need to call `window.mParticle.Rokt.getLocalSessionAttributes` if there is no eventMappingLookup

src/Rokt-Kit.js
rmi22186: Just realized that the reason is because the robot told you some wrong information when you did in fact use the code i referenced above.  I tested the robot's advice:

"The isEmpty function logic is flawed. If value is null or undefined, Object.keys(value) will throw an error before the null check can prevent it"

Null check is the first thing it does though...so when value is null or undefined it's working fine.  I'd keep it to what we had it as before, since we're using it in core/kits over the last few years.

image

rmi22186 (changes requested):

rmi22186 (approved):

PR #86, opened 2026-04-08
feat: add batch stream wrapper function

khushi1033:
## Summary
Implements the processBatch hook to support batch-level event forwarding. When called by the mParticle SDK after each batch upload, the kit forwards the full batch object (events, user attributes, identities, consent state, and non-event lifecycle data) to window.Rokt.__batch_stream__ on the Rokt WSDK. Mirrors the existing event queue pattern with two new queues: batchQueue (drains when the kit initialises) and batchStreamQueue (drains when __batch_stream__ becomes available).

Updated @mparticle/web-sdk to pull in the latest type definitions (Batch, SDKEvent, IMParticleUser, KitInterface). To fully conform to KitInterface, the following changes were made:
- init and process parameter types widened to match the interface (Record<string, unknown> / SDKEvent), with internal casts to the kit's typed shapes
- Identity callbacks (onUserIdentified, onLoginComplete, onLogoutComplete, onModifyComplete) updated to accept IMParticleUser with the correct second parameter where required
- Public interface methods (init, process, processBatch, identity callbacks, attribute methods) now return descriptive strings on success and error to comply with KitInterface
- id property added to satisfy the required KitInterface.id field
- processEventQueue renamed to drainQueues to reflect that it now drains both the event and batch pre-init queues

## Testing Plan
Added #processBatch tests in test/src/tests.spec.ts covering: happy path forwarding, non-event data passthrough, pre-init queuing and flush, __batch_stream__ unavailable queuing and flush, window.Rokt undefined graceful fallback, and flush ordering. All tests existing from before continue to pass without modification.


src/Rokt-Kit.ts
```diff
@@ -928,147 +966,161 @@ class RoktKit {
...
+      return "Can't initialize forwarder: " + name + ': ' + e;
     }
   }
 
-  public process(event: MParticleEvent): void {
-    if (!this.isKitReady()) {
-      this.eventQueue.push(event);
-      return;
-    }
+  public process(event: SDKEvent): string {
```
jamesnrokt: I think the whole implementation of events can be removed now in favor of the batches

src/Rokt-Kit.ts
```diff
@@ -1081,36 +1133,66 @@ class RoktKit {
...
 
-  public onUserIdentified(filteredUser: FilteredUser): void {
-    this.filters.filteredUser = filteredUser;
-    this.userAttributes = filteredUser.getAllUserAttributes();
-    this.sendEventStream(this.buildIdentityEvent('identify', filteredUser));
+  public onUserIdentified(user: IMParticleUser): string {
+    try {
+      this.filters.filteredUser = user as FilteredUser;
+      this.userAttributes = user.getAllUserAttributes();
+      this.sendEventStream(this.buildIdentityEvent('identify', user as FilteredUser));
```
jamesnrokt: Any "special" events like this conform to the batch schema and should be added to the next batch before we send it to wSDK

jamesnrokt (commented):
Fairly hard to review given the changes to implement the interface at the same time as the batching updates. May be worth splitting?

src/Rokt-Kit.ts
```diff
@@ -72,12 +74,12 @@ interface RoktGlobal {
   createLocalLauncher(options: Record<string, unknown>): RoktLauncher;
   currentLauncher?: RoktLauncher;
   __event_stream__?(event: Record<string, unknown>): void;
+  __batch_stream__?(batch: Batch): void;
```
alexs-mparticle, line 78: We can also remove `__event_stream__`

src/Rokt-Kit.ts
```diff
@@ -72,12 +74,12 @@ interface RoktGlobal {
...
   setExtensionData(data: Record<string, unknown>): void;
 }
 
-interface FilteredUser {
-  getMPID(): string;
-  getAllUserAttributes(): Record<string, unknown>;
+interface FilteredUser extends IMParticleUser {
+  getMPID?(): string;
```
alexs-mparticle: Why are we removing `getAllUserAttributes()` but keeping the others?

alexs-mparticle: Also, I don't think `getMPID` would ever be optional on a user.

src/Rokt-Kit.ts
```diff
@@ -760,11 +765,16 @@ class RoktKit {
     mp().logEvent(EVENT_NAME_SELECT_PLACEMENTS, EVENT_TYPE_OTHER, attributes as Record<string, unknown>);
   }
 
-  private processEventQueue(): void {
+  private drainQueues(): void {
     this.eventQueue.forEach((event) => {
       this.process(event);
     });
     this.eventQueue = [];
+
```
alexs-mparticle, line 772: We probably should remove the eventQueue logic too.

alexs-mparticle (changes requested):

src/Rokt-Kit.ts
rmi22186: We shouldn't need to extend `IMParticleUser` at all.  if you look at the definition, it extends, [User](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/mparticle__web-sdk/index.d.ts#L637-L658), which has all of these.  

Yes it's technically a filtered user, but there's no difference bewteen an IMparticleUser and a FilteredUser from an interface perspective, so let's just use `IMparticleUser`.

src/Rokt-Kit.ts
```diff
@@ -1042,34 +1077,39 @@ class RoktKit {
       target.appendChild(script);
       this.captureTiming(RoktKit.PERFORMANCE_MARKS.RoktScriptAppended);
     }
+
+    return 'Successfully initialized: ' + name;
   }
 
-  public process(event: MParticleEvent): void {
```
rmi22186, line 1047: now that you are importing `SDKEvent`.  You can remove `MParticleEvent` from everywhere in this file and replace it with `SDKEvent` and also remove the `MParticleEvent` interface

src/Rokt-Kit.ts
```diff
@@ -1042,34 +1077,39 @@ class RoktKit {
...
+  public process(event: SDKEvent): string {
     if (!this.isKitReady()) {
       this.eventQueue.push(event);
-      return;
+      return 'Event queued for forwarder: ' + name;
     }
+    const mpEvent = event as unknown as MParticleEvent;
```
rmi22186: Once you remove `MParticleEvent` from everywhere you should be able to remove this line and rever mpEvent back to event in the below code.

src/Rokt-Kit.ts
alexs-mparticle: Works for me!

test/src/tests.spec.ts
```diff
@@ -5149,6 +5149,239 @@ describe('Rokt Forwarder', () => {
     });
   });
 
+  describe('#processBatch', () => {
+    let mockBatch: any;
```
rmi22186: may be worth updating this from any to `Batch` which is imported now.  If it causes too many issues in the tests below, we can make a clean up ticket for the future

src/Rokt-Kit.ts
khushi1033: Yep removed!

src/Rokt-Kit.ts
khushi1033: Done!

src/Rokt-Kit.ts
khushi1033: Fixed no longer optional

src/Rokt-Kit.ts
khushi1033: Done!

test/src/tests.spec.ts
```diff
@@ -5149,6 +5149,239 @@ describe('Rokt Forwarder', () => {
...
+      (window as any).mParticle.forwarder.processBatch(mockBatch);
+
+      expect(receivedBatches.length).toBe(1);
+      expect(receivedBatches[0].mpid).toBe('test-mpid-123');
+      expect(receivedBatches[0].user_attributes).toEqual({ 'user-attr': 'user-value' });
+      expect(receivedBatches[0].user_identities).toEqual({ email: 'test@example.com' });
+      expect(receivedBatches[0].events.length).toBe(1);
+    });
+
+    it('should forward batch including non-event types (UserIdentityChange, UserAttributeChange)', async () => {
```
rmi22186: nit: `non-event` types is awkward here. 

These aren't "non-events" reallly.  They are events that when they are in the form of SDKEvent (prior to being converted into a structure that adheres to the JSON schema), have an `EventName` of a number, and the final event once it is converted doesn't have an `event_name`, but instead, an `event_type`.  I don't think there is value in this test. Remove...or ask claude and see what it says :)

src/Rokt-Kit.ts
rmi22186: @khushi1033 - just making sure yous ee this comment, since you pushed a change to fix alex's update, which made this comment, and my reply "outdated", and collapsed the comment.

khushi1033: @rmi22186 Yeah sorry was still looking into it -  I'm running into an issue though, since `getMPID` and `getUserIdentities` live on the User base type, which IMParticleUser extends, Typescript can't see them. Seeing [in this comment](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/public-types.ts#L10-L14)  it's intentionally  not re-exported so thinking might have to stick to extending until that's updated?

test/src/tests.spec.ts
khushi1033: Yep removed👍

test/src/tests.spec.ts
khushi1033: Actually no issues done!

src/Rokt-Kit.ts
khushi1033: Done!

src/Rokt-Kit.ts
khushi1033: Done!

src/Rokt-Kit.ts
khushi1033: Yep done!

alexs-mparticle (approved):

src/Rokt-Kit.ts
jamesnrokt: Before a batch leaves the kit any events from: `onModifyComplete`, `onLogoutComplete`, `onLoginComplete` or `onUserIdentified` need to be added to a batch.

So you'll still have 2 arrays, batches to be sent and non-batch events from those above methods to be added to a batch

src/Rokt-Kit.ts
rmi22186: @khushi1033 that's probably fine for now because eventually will be removed from .  I'd add a comment saying that it needs to be cleaned up in the future so we don't forget.  Something to add to our cool down week I'd say.

src/Rokt-Kit.ts
khushi1033: Ah ok, sorry I though you were referring to removing sending events now that we're using batches.  

That makes sense though - implemented it so those events are now pushed onto a `pendingIdentityEvents` array. When processBatch is called, `mergePendingIdentityEvents` appends any pending events to the batch's events array and clears the queue before the batch is sent.

src/Rokt-Kit.ts
```diff
@@ -203,7 +200,7 @@ const moduleId = 181;
 const EVENT_NAME_SELECT_PLACEMENTS = 'selectPlacements';
 const ADBLOCK_CONTROL_DOMAIN = 'apps.roktecommerce.com';
 const INIT_LOG_SAMPLING_RATE = 0.1;
-const MESSAGE_TYPE_PROFILE = 14; // mParticle MessageType.Profile
+const MESSAGE_TYPE_PROFILE = 14;
```
jamesnrokt:
```suggestion
const MESSAGE_TYPE_PROFILE = 14; // mParticle MessageType.Profile
```

Not sure if `MessageType.Profile` is now included in our imported types. If not then we should at least re-instate the comment

jamesnrokt (approved):

rmi22186 (approved):

PR #89, opened 2026-04-21
feat: Embed Thank You Element in Joint SDK

zach-will:
## Background
Today, partners need to load Rokt resources independently in order to use shoppable ads. This has introduced confusion and added integration complexity. To simplify this, the Joint SDK will now be responsible for fetching the Thank You Element (TYE).

## What Changed
- Rokt Kit now handles the `thank-you-journey` extension as a special case. When this extension is present, it is no longer appended to the Rokt launcher URL. Instead, Rokt kit will fetch the TYE resource and register the extension with WSDK.
- `mParticle.rokt.use` has been deprecated, as this functionality is now handled internally.
- Updated TsConfig to fix local development issues with vitest

## Testing and Verification
- Conducted local test using stub config data.

## Media
TODO

## Expected config:

```
      "roktExtensions": [
        {"jsmap": null, "map": null, "maptype": "StaticList", "value": "instant-purchase"},
        {"jsmap": null, "map": null, "maptype": "StaticList", "value": "experiment-monitoring"},
        {"jsmap": null, "map": null, "maptype": "StaticList", "value": "thank-you-journey"}
      ],
```

### DIFF
--- a/src/Rokt-Kit.ts
+++ b/src/Rokt-Kit.ts
@@ -178,6 +178,11 @@ interface LogEntry {
   code?: string;
 }
 
+interface RoktExtensionConfig {
+  roktExtensionsQueryParams: string[];
+  legacyRoktExtensions: string[];
+}
+
 declare global {
   interface Window {
     Rokt?: RoktGlobal;
@@ -206,6 +211,7 @@ const ROKT_IDENTITY_EVENT_TYPE = {
   MODIFY_USER: 'modify_user',
   IDENTIFY: 'identify',
 } as const;
+const ROKT_THANK_YOU_JOURNEY_EXTENSION = 'ThankYouJourney';
 
 type RoktIdentityEventType = (typeof ROKT_IDENTITY_EVENT_TYPE)[keyof typeof ROKT_IDENTITY_EVENT_TYPE];
 
@@ -245,15 +251,25 @@ function mp(): MParticleExtended {
 // ============================================================
 
 function generateLauncherScript(domain: string | undefined, extensions: string[]): string {
-  const resolvedDomain = typeof domain !== 'undefined' ? domain : 'apps.rokt.com';
-  const protocol = 'https://';
   const launcherPath = '/wsdk/integrations/launcher.js';
-  const baseUrl = [protocol, resolvedDomain, launcherPath].join('');
+  const baseUrl = [generateBaseUrl(domain), launcherPath].join('');
 
   if (!extensions || extensions.length === 0) {
     return baseUrl;
   }
   return baseUrl + '?extensions=' + extensions.join(',');
+}
+
+function generateThankYouElementScript(domain: string | undefined) {
+  const thankYouElementPath = '/rokt-elements/rokt-element-thank-you.js';
+  return [generateBaseUrl(domain), thankYouElementPath].join('');
+}
+
+function generateBaseUrl(domain: string |undefined) {
+  const resolvedDomain = typeof domain !== 'undefined' ? domain : 'apps.rokt.com';
+  const protocol = 'https://';
+
+  return [protocol, resolvedDomain].join('');
 }
 
 function isObject(val: unknown): val is Record<string, unknown> {
@@ -272,6 +288,7 @@ function parseSettingsString<T>(settingsString?: string): T[] {
   return [];
 }
 
+// TODO Remove and fix test
 function extractRoktExtensions(settingsString?: string): string[] {
   const settings = settingsString ? parseSettingsString<RoktExtensionEntry>(settingsString) : [];
   const roktExtensions: string[] = [];
@@ -281,6 +298,47 @@ function extractRoktExtensions(settingsString?: string): string[] {
   }
 
   return roktExtensions;
+}
+
+function extractRoktExtensionConfig(domain?: string, settingsString?: string): RoktExtensionConfig {
+  const settings = settingsString ? parseSettingsString<RoktExtensionEntry>(settingsString) : [];
+  const roktExtensionsQueryParams: string[] = [];
+  const legacyRoktExtensions: string[] = [];
+
+  for (let i = 0; i < settings.length; i++) {
+    const extensionName = settings[i].value;
+    if (extensionName === 'thank-you-journey') {
+      loadRoktThankYouElement(domain);
+      legacyRoktExtensions.push(ROKT_THANK_YOU_JOURNEY_EXTENSION)
+    } else {
+      roktExtensionsQueryParams.push(settings[i].value);
+    }
+  }
+
+  return { roktExtensionsQueryParams, legacyRoktExtensions };
+}
+
+function loadRoktThankYouElement(domain?: string) {
+  const scriptId = 'rokt-thank-you-element';
+  if (document.getElementById(scriptId)) {
+    return;
+  }
+
+  const target = document.head || document.body
+  const script = document.createElement('script');
+  script.type = 'text/javascript';
+  (script as HTMLScriptElement & { fetchPriority: string }).fetchPriority = 'high';
+  script.src = generateThankYouElementScript(domain);
+  script.crossOrigin = 'anonymous';
+  script.async = true;
+  script.id = scriptId;
+  target.appendChild(script)
+}
+
+function registerLegacyExtensions(legacyExtensions: string[]) {
+  for (const extension of legacyExtensions) {
+    window.mParticle.Rokt.use(extension);
+  }
 }
 
 function generateMappedEventLookup(placementEventMapping: PlacementEventMappingEntry[]): Record<string, string> {
@@ -954,7 +1012,6 @@ class RoktKit implements KitInterface {
   ): string {
     const kitSettings = settings as unknown as RoktKitSettings;
     const accountId = kitSettings.accountId;
-    const roktExtensions = extractRoktExtensions(kitSettings.roktExtensions);
     this.userAttributes = filteredUserAttributes || {};
     this._onboardingExpProvider = kitSettings.onboardingExpProvider;
 
@@ -972,6 +1029,7 @@ class RoktKit implements KitInterface {
     }
 
     const domain = mp().Rokt?.domain;
+    const { roktExtensionsQueryParams, legacyRoktExtensions } = extractRoktExtensionConfig(domain, kitSettings.roktExtensions);
     const launcherOptions: Record<string, unknown> = {
       ...((mp().Rokt?.launcherOptions as Record<string, unknown>) || {}),
     };
@@ -1040,7 +1098,7 @@ class RoktKit implements KitInterface {
       const target = document.head || document.body;
       const script = document.createElement('script');
       script.type = 'text/javascript';
-      script.src = generateLauncherScript(domain, roktExtensions);
+      script.src = generateLauncherScript(domain, roktExtensionsQueryParams);
       script.async = true;
       script.crossOrigin = 'anonymous';
       (script as HTMLScriptElement & { fetchPriority: string }).fetchPriority = 'high';
@@ -1049,6 +1107,7 @@ class RoktKit implements KitInterface {
       script.onload = () => {
         if (this.isLauncherReadyToAttach()) {
           this.attachLauncher(accountId, launcherOptions);
+          registerLegacyExtensions(legacyRoktExtensions);
         } else {
           console.error('Rokt object is not available after script load.');
         }
--- a/src/Rokt-Kit.ts
+++ b/src/Rokt-Kit.ts
@@ -277,25 +277,45 @@ function generateBaseUrl(domain: string | undefined) {
   return [protocol, resolvedDomain].join('');
 }
 
-function loadRoktScript(scriptId: string, source: string, appendToTarget: boolean = true) {
-  const preexistingScript = document.getElementById(scriptId);
-  if (preexistingScript) {
-    return preexistingScript;
-  }
+//function loadRoktScript(scriptId: string, source: string, appendToTarget: boolean = true) {
+//  const preexistingScript = document.getElementById(scriptId);
+//  if (preexistingScript) {
+//    return preexistingScript;
+//  }
+
+//  const target = document.head || document.body;
+//  const script = document.createElement('script');
+//  script.type = 'text/javascript';
+//  (script as HTMLScriptElement & { fetchPriority: string }).fetchPriority = 'high';
+//  script.src = source;
+//  script.crossOrigin = 'anonymous';
+//  script.async = true;
+//  script.id = scriptId;
+//  if (appendToTarget) {
+//    target.appendChild(script);
+//  }
+
+//  return script;
+//}
+
+function loadRoktScript(
+  scriptId: string,
+  source: string,
+  handlers?: { onLoad?: () => void; onError?: (e: Event | string) => void },
+): void {
+  if (document.getElementById(scriptId)) return; // resolves the preexisting script issue
 
   const target = document.head || document.body;
   const script = document.createElement('script');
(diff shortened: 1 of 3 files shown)


src/Rokt-Kit.ts
```diff
@@ -272,6 +288,7 @@ function parseSettingsString<T>(settingsString?: string): T[] {
   return [];
 }
 
+// TODO Remove and fix test
```
rmi22186: Can this TODO be removed?

src/Rokt-Kit.ts
```diff
@@ -283,6 +300,47 @@ function extractRoktExtensions(settingsString?: string): string[] {
...
+  const roktExtensionsQueryParams: string[] = [];
+  const legacyRoktExtensions: string[] = [];
+
+  for (let i = 0; i < settings.length; i++) {
+    const extensionName = settings[i].value;
+    if (extensionName === 'thank-you-journey') {
+      loadRoktThankYouElement(domain);
+      legacyRoktExtensions.push(ROKT_THANK_YOU_JOURNEY_EXTENSION)
+    } else {
+      roktExtensionsQueryParams.push(settings[i].value);
```
rmi22186:
```suggestion
      roktExtensionsQueryParams.push(extensionName);
```

src/Rokt-Kit.ts
```diff
@@ -283,6 +300,47 @@ function extractRoktExtensions(settingsString?: string): string[] {
...
+
+  const target = document.head || document.body
+  const script = document.createElement('script');
+  script.type = 'text/javascript';
+  (script as HTMLScriptElement & { fetchPriority: string }).fetchPriority = 'high';
+  script.src = generateThankYouElementScript(domain);
+  script.crossOrigin = 'anonymous';
+  script.async = true;
+  script.id = scriptId;
+  target.appendChild(script)
```
rmi22186: nit: Let's DRY this up with the other script tag creation for launcher.js

src/Rokt-Kit.ts
```diff
@@ -283,6 +300,47 @@ function extractRoktExtensions(settingsString?: string): string[] {
...
+}
+
+function registerLegacyExtensions(legacyExtensions: string[]) {
+  for (const extension of legacyExtensions) {
+    window.mParticle.Rokt.use(extension);
+  }
+}
```
rmi22186, line 343: Is this necessary? I thought we are doing this in the new script creation.

rmi22186 (changes requested):
Looking good! Can you add a few tests also?

src/Rokt-Kit.ts
```diff
@@ -283,6 +300,47 @@ function extractRoktExtensions(settingsString?: string): string[] {
...
 
+function extractRoktExtensionConfig(domain?: string, settingsString?: string): RoktExtensionConfig {
+  const settings = settingsString ? parseSettingsString<RoktExtensionEntry>(settingsString) : [];
+  const roktExtensionsQueryParams: string[] = [];
+  const legacyRoktExtensions: string[] = [];
+
+  for (let i = 0; i < settings.length; i++) {
+    const extensionName = settings[i].value;
+    if (extensionName === 'thank-you-journey') {
+      loadRoktThankYouElement(domain);
```
rmi22186: Since the method name is `extractRoktExtensionConfig`, having it also load the TYE is a bit unexpected. Can you move that to the init, parse returns?

src/Rokt-Kit.ts
```diff
@@ -245,17 +251,27 @@ function mp(): MParticleExtended {
...
   }
   return baseUrl + '?extensions=' + extensions.join(',');
 }
 
+function generateThankYouElementScript(domain: string | undefined) {
+  const thankYouElementPath = '/rokt-elements/rokt-element-thank-you.js';
+  return [generateBaseUrl(domain), thankYouElementPath].join('');
+}
+
+function generateBaseUrl(domain: string |undefined) {
```
rmi22186:
```suggestion
function generateBaseUrl(domain: string | undefined) {
```

rmi22186: Actually, will this ever return `undefined` if we are generating just the base portion of a URL?

src/Rokt-Kit.ts
```diff
@@ -245,17 +251,27 @@ function mp(): MParticleExtended {
...
-  const baseUrl = [protocol, resolvedDomain, launcherPath].join('');
+  const baseUrl = [generateBaseUrl(domain), launcherPath].join('');
 
   if (!extensions || extensions.length === 0) {
     return baseUrl;
   }
   return baseUrl + '?extensions=' + extensions.join(',');
 }
 
+function generateThankYouElementScript(domain: string | undefined) {
```
rmi22186, line 268: Will this ever be undefined?

zach-will: I believe it will be if partners don't define `ROKT_DOMAIN`.

src/Rokt-Kit.ts
zach-will: Updated

src/Rokt-Kit.ts
zach-will: Updated.

src/Rokt-Kit.ts
zach-will: Updated

src/Rokt-Kit.ts
zach-will: `generateBaseUrl` will never return undefined. I believe parameter needs to have the union to support cname?

src/Rokt-Kit.ts
zach-will: Yup. This is necessary to initialize the WSDK thank you page extension manager.

src/Rokt-Kit.ts
zach-will: Updated to use `loadRoktScript`

src/Rokt-Kit.ts
rmi22186: From our [docs](https://docs.rokt.com/developers/integration-guides/getting-started/ecommerce/ecommerce-sdk-integration/#sdk-initialization-script), if you search for ROKT_DOMAIN, you'll see we set it as follows:
```
window.ROKT_DOMAIN = ROKT_DOMAIN || 'https://apps.rokt-api.com';
```

I think it's safe to keep here, not a big deal, more of an FYI.

src/Rokt-Kit.ts
rmi22186: Ah, right!

src/Rokt-Kit.ts
```diff
@@ -265,13 +269,34 @@ function generateThankYouElementScript(domain: string | undefined) {
...
 
-function generateBaseUrl(domain: string |undefined) {
+function generateBaseUrl(domain: string | undefined) {
   const resolvedDomain = typeof domain !== 'undefined' ? domain : 'apps.rokt.com';
   const protocol = 'https://';
 
   return [protocol, resolvedDomain].join('');
 }
 
+function loadRoktScript(scriptId: string, source: string, appendToTarget: boolean = true) {
```
rmi22186: I think this should be a void, there isn't a need to return anything, unless I'm mistaken

src/Rokt-Kit.ts
```diff
@@ -245,17 +255,48 @@ function mp(): MParticleExtended {
...
+  const resolvedDomain = typeof domain !== 'undefined' ? domain : 'apps.rokt.com';
+  const protocol = 'https://';
+
+  return [protocol, resolvedDomain].join('');
+}
+
+function loadRoktScript(scriptId: string, source: string, appendToTarget: boolean = true) {
+  const preexistingScript = document.getElementById(scriptId);
+  if (preexistingScript) {
+    return preexistingScript;
```
rmi22186:
```suggestion
    return;
```

src/Rokt-Kit.ts
```diff
@@ -245,17 +255,48 @@ function mp(): MParticleExtended {
...
+  (script as HTMLScriptElement & { fetchPriority: string }).fetchPriority = 'high';
+  script.src = source;
+  script.crossOrigin = 'anonymous';
+  script.async = true;
+  script.id = scriptId;
+  if (appendToTarget) {
+    target.appendChild(script);
+  }
+  
+  return script;
```
rmi22186:
```suggestion
```

src/Rokt-Kit.ts
```diff
@@ -245,17 +255,48 @@ function mp(): MParticleExtended {
...
   return baseUrl + '?extensions=' + extensions.join(',');
 }
 
+function generateThankYouElementScript(domain: string | undefined) {
+  const thankYouElementPath = '/rokt-elements/rokt-element-thank-you.js';
+  return [generateBaseUrl(domain), thankYouElementPath].join('');
+}
+
+function generateBaseUrl(domain: string | undefined) {
+  const resolvedDomain = typeof domain !== 'undefined' ? domain : 'apps.rokt.com';
```
rmi22186: The default domain for joint sdk is `apps.rokt-api.com`
```suggestion
  const resolvedDomain = typeof domain !== 'undefined' ? domain : 'apps.rokt-api.com';
```

src/Rokt-Kit.ts
```diff
@@ -272,15 +313,33 @@ function parseSettingsString<T>(settingsString?: string): T[] {
...
+  return {
+    roktExtensionsQueryParams,
+    legacyRoktExtensions,
+    loadThankYouElement,
+   };
+}
+
+function registerLegacyExtensions(legacyExtensions: string[]) {
+  for (const extension of legacyExtensions) {
+    window.mParticle.Rokt.use(extension);
```
rmi22186: Up for discussion - I think we should just do 
```suggestion
    window.Rokt.use(extension);
```
instead. `window.mParticle.Rokt.use` is the public API for someone consuming the mParticle SDK. In the Rokt Kit, we make calls directly to Rokt.  Especially if we are going to deprecate window.mParticle.Rokt.use per the below warning.

rmi22186: Actually - question about the flow and usage.  If the SDK is loaded on the home screen, and .use is called on the home screen, is that okay?  I'm not familiar enough with TYE to understand when it shows. Is it when .use is called?  Or when `selectPlacements` is called?

test/src/tests.spec.ts
```diff
@@ -3272,38 +3272,139 @@ describe('Rokt Forwarder', () => {
     });
   });
 
-  describe('#roktExtensions', () => {
-    beforeEach(async () => {
-      (window as any).Rokt = new (MockRoktForwarder as any)();
-      (window as any).mParticle.Rokt = (window as any).Rokt;
+  describe('#generateThankYouElementScript', () => {
+    const baseUrl = 'https://apps.rokt.com/rokt-elements/rokt-element-thank-you.js';
```
rmi22186:
```suggestion
    const baseUrl = 'https://apps.rokt-api.com/rokt-elements/rokt-element-thank-you.js';
```

test/src/tests.spec.ts
```diff
@@ -3272,38 +3272,139 @@ describe('Rokt Forwarder', () => {
...
+        false,
+      );
 
-      expect((window as any).mParticle.forwarder.testHelpers.extractRoktExtensions(undefined)).toEqual([]);
+      const tyeScript = document.getElementById('rokt-thank-you-element') as HTMLScriptElement;
+      expect(tyeScript).not.toBeNull();
+      expect(tyeScript.src).toContain('/rokt-elements/rokt-element-thank-you.js');
+    });
+
+    it('should call mParticle.Rokt.use with ThankYouJourney when thank-you-journey extension is provided', async () => {
```
rmi22186: will have to be updated to use Rokt instead of mParticle.Rokt if the Rokt.use is implemented based on prior feedback

rmi22186 (changes requested):
a few more comments

src/Rokt-Kit.ts
```diff
@@ -272,15 +313,33 @@ function parseSettingsString<T>(settingsString?: string): T[] {
...
+      roktExtensionsQueryParams.push(extensionName);
+    }
   }
 
-  return roktExtensions;
+  return {
+    roktExtensionsQueryParams,
+    legacyRoktExtensions,
+    loadThankYouElement,
+   };
```
rmi22186: run a `npm run lint:fix.` to catch anything else
```suggestion
  };
```

src/Rokt-Kit.ts
zach-will: We return the script tag because we handle appending the script to the document differently for TYE and WSDK. For TYE we append the script to document as soon as the script is created, for WSDK we append script to document once the script has finished loading.

I wasn't sure if there was a reason why WSDK script is appended to the DOM much later, but I didn't want to break any behavior that could be related.

src/Rokt-Kit.ts
zach-will: Updated

test/src/tests.spec.ts
zach-will: Updated

src/Rokt-Kit.ts
zach-will: Same reason as the comment above. If there's no reason to append WSDK to document after script load, I can remove the return from the script loader and standardize appending to doc.

src/Rokt-Kit.ts
zach-will: Updated

src/Rokt-Kit.ts
zach-will: https://github.com/mparticle-integrations/mparticle-javascript-integration-rokt/pull/89#discussion_r3124762971

test/src/tests.spec.ts
zach-will: Updated

src/Rokt-Kit.ts
zach-will: Updated to use Window.Rokt.use.

Yeah this is ok. `ThankYouPageExtensionManager` in WSDK will analyze selections from `selectPlacements` and if no selection contains `target-section: thanks` WSDK will bypass TYE.

`thank-you-element.js` defines the custom element `<rokt-thank-you>`. Given `<rokt-thank-you>` is present in the DOM and the script loads, it immediately evaluates visibility eligibility and shows a loading state.

After selectPlacements resolves, TYE transitions either to shoppable ads or to a confirmation page, depending on whether a selection exists with `target-section: "thanks"`.

rmi22186: Makes sense!

src/Rokt-Kit.ts
rmi22186: I see now.  There's different behavior happening when creating scripts for TYE vs Rokt SDK because of onLoad/onError.  The

how about something like:

```
function loadRoktScript(
    scriptId: string,
    source: string,
    handlers?: { onload?: () => void; onerror?: (e: Event | string) => void },
  ): void {
    if (document.getElementById(scriptId)) return; // resolves the preexisting script issue

    const target = document.head || document.body;
    const script = document.createElement('script');
    script.id = scriptId;
    script.type = 'text/javascript';
    script.src = source;
    script.async = true;
    script.crossOrigin = 'anonymous';
    (script as HTMLScriptElement & { fetchPriority: string }).fetchPriority = 'high';
    if (handlers?.onload) script.onload = handlers.onload;
    if (handlers?.onerror) script.onerror = handlers.onerror;
    target.appendChild(script);
  }
```

rmi22186: It has to delay appending due to having the onload/onerror

zach-will: Updated

rmi22186 (approved):
LGTM. Will do some testing first to make sure existing script loading is unaffected. Then will merge and release tomorrow

PR #91, opened 2026-04-23
Fix/Enable Thank You Element Partner Event listening before RoktKit is ready

zach-will:
## Background
As part of embedding the Rokt Thank You Element (TYE) into the Joint SDK, we need a way for partners to register callbacks and listen to TYE events.

Previously, partners were instructed to attach event listeners inside the TYE onLoad handler:

```
script.src = baseUrl + '/rokt-thank-you-element.js'
script.onLoad = () => {
  window.RoktThankYouElement.on('THANK_YOU_ELEMENT_LOADING_INITIATED').subscribe(() => {
    // Do something
  });
}
```

With TYE now internalized within the Joint SDK, this approach is no longer viable.

We also want to allow partners to register event listeners before mParticle or RoktKit is fully ready. To support this, we’ve introduced a new method, `onShoppableAdsReady`, which exposes a safe hook for registering TYE event listeners.

The updated integration pattern is:

```
  window.mParticle.Rokt.onShoppableAdsReady(() => {
    window.RoktThankYouElement.on('THANK_YOU_ELEMENT_LOADING_INITIATED').subscribe(() => {
      // Do something
    });
  });
  window.mParticle?.ready?.(() => {
```

## What Changed
- Updated internal extension registration:
  - mParticle.rokt.use → launcher.use
- Added a new public method: onShoppableAdsReady in RoktKit
- When ThankYouElement is included in the kit config:
  - RoktKit is initialized early via RoktManager, but is not marked as ready until the launcher is defined
  - Just before loading rokt-thank-you-element.js, the onShoppableAdsReady callback queue in RoktManager is flushed
    - This ensures all queued callbacks are registered as onLoad handlers
    - Any subsequent calls to onShoppableAdsReady are executed immediately (no longer deferred)

## Changes to snippet js.
Integrations will now require `onShoppableAdsReady` to be included in the RoktMethods stub:

```
[
      'selectPlacements',
      'hashAttributes',
      'setExtensionData',
      'use',
      'onShoppableAdsReady',
    ].forEach(
    (name) => {
      window.mParticle.Rokt[name] = (...args: any[]) =>
        window.mParticle.config.rq.push([`Rokt.${name}`, ...args])
    },
  )
```

## Deployment order
- Deploy RoktManager Changes [PR-1254](https://github.com/mParticle/mparticle-web-sdk/pull/1254)
- Deploy this PR

## Media

https://github.com/user-attachments/assets/62203695-3c4b-4918-8abb-9a737372854c


src/Rokt-Kit.ts
```diff
@@ -336,9 +336,11 @@ function extractRoktExtensionConfig(settingsString?: string): RoktExtensionConfi
   };
 }
 
-function registerLegacyExtensions(legacyExtensions: string[]) {
-  for (const extension of legacyExtensions) {
-    window.Rokt?.use(extension);
+function registerLegacyExtensions(legacyExtensions: string[], launcher: RoktLauncher | null) {
+  if (launcher) {
+    for (const extension of legacyExtensions) {
+      launcher.use(extension);
```
rmi22186: Per https://docs.rokt.com/developers/integration-guides/shoppable-ads/sdk-integration-guides/new-web-sdk-integration/#step-4-select-a-placement docs - should this be an `await`?

test/src/tests.spec.ts
```diff
@@ -3375,17 +3382,98 @@ describe('Rokt Forwarder', () => {
...
+          filterUserAttributes: (attrs: any) => attrs,
+          filteredUser: { getMPID: () => '123' },
+        },
+        use: () => Promise.resolve(),
+      };
+
+      await (window as any).mParticle.forwarder.init(
+        {
+          accountId: '123456',
+          roktExtensions: '[{"jsmap":null,"map":null,"maptype":"LegacyExtension","value":"thank-you-journey"}]',
```
rmi22186: Not that it matters, but there isn't a maptype of `LegacyExtension`...let's make it staticlist to better match what the server sends back
```suggestion
          roktExtensions: '[{"jsmap":null,"map":null,"maptype":"StaticList","value":"thank-you-journey"}]',
```

test/src/tests.spec.ts
```diff
@@ -3375,17 +3382,98 @@ describe('Rokt Forwarder', () => {
...
+          userAttributesFilters: [],
+          filterUserAttributes: (attrs: any) => attrs,
+          filteredUser: { getMPID: () => '123' },
+        },
+      };
+
+      await (window as any).mParticle.forwarder.init(
+        {
+          accountId: '123456',
+          roktExtensions: '[{"jsmap":null,"map":null,"maptype":"LegacyExtension","value":"thank-you-journey"}]',
```
rmi22186:
```suggestion
          roktExtensions: '[{"jsmap":null,"map":null,"maptype":"StaticList","value":"thank-you-journey"}]',
```

src/Rokt-Kit.ts
zach-will: Updated to use Promise.all. Callers now use then or await.

test/src/tests.spec.ts
zach-will: removed everywhere

test/src/tests.spec.ts
zach-will: removed everywhere

src/Rokt-Kit.ts
```diff
@@ -77,7 +77,6 @@ interface RoktGlobal {
   currentLauncher?: RoktLauncher;
   __batch_stream__?(batch: Batch): void;
   setExtensionData(data: Record<string, unknown>): void;
-  use(value: string): void;
```
alexs-mparticle, line 80: Why are we removing this method? This feels like an accidental delete.

src/Rokt-Kit.ts
```diff
@@ -95,6 +94,7 @@ interface KitFilters {
 
 interface RoktManager {
   attachKit(kit: RoktKit): void | Promise<void>;
+  flushOnShoppableAdsReadyMessageQueue(kit: RoktKit): void;
```
alexs-mparticle: I think we should make this optional since it appears to be dependent on the "thank you" element, which may not always be available.

```suggestion
  flushOnShoppableAdsReadyMessageQueue?(kit: RoktKit): void;
```

alexs-mparticle (changes requested):

src/Rokt-Kit.ts
zach-will: use shouldn't be declared on the `Rokt` model. It's a method of the `RoktLauncher`.

src/Rokt-Kit.ts
zach-will: Updated.

src/Rokt-Kit.ts
alexs-mparticle: My mistake.

alexs-mparticle (approved):

PR #92, opened 2026-04-28
feat: Add Workspace IDSync search on user identification

rmi22186:
## Summary

- Adds a new optional kit setting **`workspaceIdSyncApiKey`** (a workspace's mParticle API key). When set, the kit calls `mParticle.Identity.search(apiKey, { email }, callback)` from `onUserIdentified` and, on a 200, sets a kit-local flag **`userIdentifiedInWorkspace = true`** that flows into the next `selectPlacements` call.
- Defensive paths: setting absent or empty → no network call; `getUserIdentities().userIdentities.email` missing or non-string → no call; `Identity.search` not available on the SDK → graceful no-op; a throw is swallowed and logged; 404 or any non-200 leaves the flag unset.
- The setting is a string (the workspace API key) rather than a boolean, so a single SDK install can run searches against an arbitrary workspace without reconfiguring the host page's mParticle SDK.
- Pairs with the mParticle Web SDK PR — https://github.com/mParticle/mparticle-web-sdk/pull/1255 (which exposes `mParticle.Identity.search`) — and depends on the Fastly `/v1/search` CORS allowlist being updated to include `x-mp-key`. The origin server already returns 200/404 via curl; only the browser preflight is blocked today.
- Test plan: `npm run lint` + `npm run build` clean (IIFE + CJS + ESM + type defs); `npm test` covers the 200 success path, 404 no-op, missing/empty api key, missing email identity, missing SDK method, and swallowed throws. Live browser smoke test still blocked on the Fastly allowlist update.

### DIFF
--- a/src/Rokt-Kit.ts
+++ b/src/Rokt-Kit.ts
@@ -87,6 +87,10 @@ interface FilteredUser extends IMParticleUser {
   getUserIdentities?: () => { userIdentities: Record<string, string> };
 }
 
+// TODO: Replace with `IIdentitySearchResult` from `@mparticle/web-sdk` once
+// a version that exports it is published (currently on a feature branch in
+// mParticle/mparticle-web-sdk PR #1255). The shape below is intentionally
+// structurally identical so the swap is a one-line import change.
 interface WorkspaceIdSyncResult {
   httpCode: number;
   body?: {
@@ -98,6 +102,8 @@ interface WorkspaceIdSyncResult {
   };
 }
 
+// TODO: Replace with `IdentitySearchCallback`-compatible reference from
+// `@mparticle/web-sdk` once published (mirrors `SDKIdentityApi.search`).
 type WorkspaceIdSyncSearcher = (
   apiKey: string,
   knownIdentities: { email: string },
@@ -237,6 +243,7 @@ const ROKT_THANK_YOU_JOURNEY_EXTENSION = 'ThankYouPageJourney';
 const ROKT_INTEGRATION_SCRIPT_ID = 'rokt-launcher';
 const ROKT_THANK_YOU_ELEMENT_SCRIPT_ID = 'rokt-thank-you-element';
 const USER_IDENTIFIED_IN_WORKSPACE_KEY = 'userIdentifiedInWorkspace';
+
 // Bound on how long selectPlacements will wait for an in-flight Workspace
 // IDSync search before proceeding without the userIdentifiedInWorkspace flag.
 // Long enough to cover the typical /v1/search round-trip (~50ms); short enough that a
@@ -715,10 +722,11 @@ class RoktKit implements KitInterface {
   private _thankYouElementOnLoadCallback: (() => void) | null = null;
   private _isThankYouElementLoaded = false;
   private _workspaceIdSyncApiKey?: string;
+
   // Held during a search dispatch so the next selectPlacements call;
   // can wait for the HTTP response before reading userIdentifiedInWorkspace;
   // — otherwise the first placement call ships without the flag.
-  private _workspaceSearchInFlight: Promise<void> | null = null;
+  private _workspaceSearchInFlightPromise: Promise<void> | null = null;
   // The email value sent in the most recent successful search
   // dispatch. If a subsequent identification arrives with the same email,
   // we skip the network call (the flag is still correct from the prior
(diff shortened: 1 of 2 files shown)


src/Rokt-Kit.ts
```diff
@@ -86,6 +87,23 @@ interface FilteredUser extends IMParticleUser {
   getUserIdentities?: () => { userIdentities: Record<string, string> };
 }
 
+interface WorkspaceIdSyncResult {
+  httpCode: number;
+  body?: {
+    context?: string | null;
+    mpid?: string;
+    matched_identities?: Record<string, string>;
```
alexs-mparticle, line 99: I could be wrong but we may have an actual type for this in core.

src/Rokt-Kit.ts
```diff
@@ -686,6 +714,16 @@ class RoktKit implements KitInterface {
   private _onboardingExpProvider?: string;
   private _thankYouElementOnLoadCallback: (() => void) | null = null;
   private _isThankYouElementLoaded = false;
+  private _workspaceIdSyncApiKey?: string;
+  // Held during a search dispatch so the next selectPlacements call;
+  // can wait for the HTTP response before reading userIdentifiedInWorkspace;
+  // — otherwise the first placement call ships without the flag.
+  private _workspaceSearchInFlight: Promise<void> | null = null;
```
alexs-mparticle: The name of this variable sounds like it should be a boolean.

```suggestion
  private _workspaceSearchInFlightPromise: Promise<void> | null = null;
```

src/Rokt-Kit.ts
```diff
@@ -217,6 +236,12 @@ const ROKT_IDENTITY_EVENT_TYPE = {
 const ROKT_THANK_YOU_JOURNEY_EXTENSION = 'ThankYouPageJourney';
 const ROKT_INTEGRATION_SCRIPT_ID = 'rokt-launcher';
 const ROKT_THANK_YOU_ELEMENT_SCRIPT_ID = 'rokt-thank-you-element';
+const USER_IDENTIFIED_IN_WORKSPACE_KEY = 'userIdentifiedInWorkspace';
+// Bound on how long selectPlacements will wait for an in-flight Workspace
```
alexs-mparticle, line 247: Can we put a space between this comment and the previous line so it reads cleaner?

src/Rokt-Kit.ts
```diff
@@ -670,6 +695,9 @@ class RoktKit implements KitInterface {
   public launcher: RoktLauncher | null = null;
   public filters: KitFilters = {};
   public userAttributes: Record<string, unknown> = {};
+  // Flag set by the Workspace IDSync flow on a 200 response. Stored on the
```
alexs-mparticle, line 705: Can we put a space between this comment and the previous line so it reads cleaner?

src/Rokt-Kit.ts
```diff
@@ -686,6 +714,16 @@ class RoktKit implements KitInterface {
   private _onboardingExpProvider?: string;
   private _thankYouElementOnLoadCallback: (() => void) | null = null;
   private _isThankYouElementLoaded = false;
+  private _workspaceIdSyncApiKey?: string;
+  // Held during a search dispatch so the next selectPlacements call;
```
alexs-mparticle, line 726: Can we put a space between this comment and the previous line so it reads cleaner?

src/Rokt-Kit.ts
```diff
@@ -1213,8 +1315,33 @@ class RoktKit implements KitInterface {
...
+   *
+   * Implementation note: this method stays non-async (returns the existing
+   * `RoktSelection | Promise<RoktSelection> | undefined` union) because
+   * `RoktSelection` has an optional `then?` member, which TS1058 rejects as
+   * ambiguously promise-like inside an async function's return type. The
+   * inner work runs in `_dispatchPlacements`; this wrapper just gates it on
+   * the in-flight search.
    */
   public selectPlacements(options: Record<string, unknown>): RoktSelection | Promise<RoktSelection> | undefined {
+    if (this._workspaceSearchInFlight) {
```
alexs-mparticle: Instead of a `.then` with a a new promise, could we use async/await?

alexs-mparticle (changes requested):

src/Rokt-Kit.ts
rmi22186: It was actually added as part of the core SDK changes. I'm going to add a follow up to remove this once it's published.

src/Rokt-Kit.ts
rmi22186: Tried a rewrite here to see how it would look and it ended up being a few different issues and so I'd like to keep it as is for now.  TS1058 error kept coming up - `operand used with the await keyword does not have a valid, callable then member`.  Sometimes there's a .then and sometimes not, so let's keep it this way for now.

rmi22186: Claude came to a final point where everything did pass, but it turned selectPlacements async from current sync, and for speed and certainty, I'd rather not potentially have any issues for live customers if we made that change.

alexs-mparticle (approved):

PR #97, opened 2026-05-21
feat: enrich commerce event action from Rokt.CommerceEventType

jaissica12:
## Summary
This PR adds commerce event type enrichment to the Rokt Kit. When the core SDK populates a `Rokt.CommerceEventType` custom flag on new commerce event types (e.g., `payment_succeeded`, `view_cart`), the Kit reads this flag and overrides `product_action.action` from `"unknown"` to the actual event name before sending the batch to `__batch_stream__`.

This is the Kit-side counterpart to the core SDK PR [#1269](https://github.com/mParticle/mparticle-web-sdk/pull/1269)
### Changes
- Added `enrichCommerceEventTypes()` private method to `RoktKit` — iterates batch events, skips non-commerce events, reads `custom_flags['Rokt.CommerceEventType']`, and overrides `product_action.action`
- Called in `processBatch()` before `sendBatchStream()` so enrichment happens just before `__batch_stream__`
- Safe to release independently — if the core SDK hasn't been updated yet, the custom flag is absent and enrichment is a no-op (`action` remains `"unknown"`)

## Testing Plan
- Added 5 unit tests in `#enrichCommerceEventTypes` describe block:
  - Replaces `action` with custom flag value for commerce events
  - Does not modify action when custom flag is absent
  - Does not modify non-commerce events (even if custom flag is present)
  - Enriches only commerce events in a mixed batch
  - Handles batch with no events
- Verified end-to-end on local playground:
  - `[Forwarder] Batch sent to Kit` shows `action: "unknown"`
  - `[Rokt Kit] After enrichment` shows `action: "payment_succeeded"`
  - MPServer batch remains `action: "unknown"` (unaffected)

### DIFF
--- a/src/Rokt-Kit.ts
+++ b/src/Rokt-Kit.ts
@@ -921,8 +921,27 @@ class RoktKit implements KitInterface {
       this.batchQueue.push(batch);
       return 'Batch queued for forwarder: ' + name;
     }
-    this.sendBatchStream(this.mergePendingIdentityEvents(batch));
+    const enrichedBatch = this.enrichCommerceEventTypes(this.mergePendingIdentityEvents(batch));
+    this.sendBatchStream(enrichedBatch);
     return 'Successfully sent batch to forwarder: ' + name;
+  }
+
+  private enrichCommerceEventTypes(batch: Batch): Batch {
+    if (!batch.events) {
+      return batch;
+    }
+    for (const event of batch.events) {
+      if (event.event_type !== 'commerce_event') continue;
+
+      const data = event.data as Record<string, unknown> | undefined;
+      if (!data) continue;
+
+      const commerceType = (data.custom_flags as Record<string, string> | undefined)?.['Rokt.CommerceEventType'];
+      if (commerceType) {
+        (data.product_action as Record<string, unknown>).action = commerceType;
+      }
+    }
+    return batch;
   }
 
   private sendBatchStream(batch: Batch): void {
--- a/src/Rokt-Kit.ts
+++ b/src/Rokt-Kit.ts
@@ -937,7 +937,7 @@ class RoktKit implements KitInterface {
       if (!data) continue;
 
       const commerceType = (data.custom_flags as Record<string, string> | undefined)?.['Rokt.CommerceEventType'];
-      if (commerceType) {
+      if (commerceType && isObject(data.product_action)) {
         (data.product_action as Record<string, unknown>).action = commerceType;
       }
     }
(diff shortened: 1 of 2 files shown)


src/Rokt-Kit.ts
```diff
@@ -921,10 +921,29 @@ class RoktKit implements KitInterface {
...
+    }
+    for (const event of batch.events) {
+      if (event.event_type !== 'commerce_event') continue;
+
+      const data = event.data as Record<string, unknown> | undefined;
+      if (!data) continue;
+
+      const commerceType = (data.custom_flags as Record<string, string> | undefined)?.['Rokt.CommerceEventType'];
+      if (commerceType) {
+        (data.product_action as Record<string, unknown>).action = commerceType;
```
rmi22186: Can throw for valid commerce events that do not have data.product_action. Core’s commerce converter allows `product_action` to be `null` while still producing `commerce_event` records with `promotion_action`, `product_impressions`. Check that data.product_action is an object using `isObject`

rmi22186 (approved):

src/Rokt-Kit.ts
```diff
@@ -921,10 +921,29 @@ class RoktKit implements KitInterface {
...
+    if (!batch.events) {
+      return batch;
+    }
+    for (const event of batch.events) {
+      if (event.event_type !== 'commerce_event') continue;
+
+      const data = event.data as Record<string, unknown> | undefined;
+      if (!data) continue;
+
+      const commerceType = (data.custom_flags as Record<string, string> | undefined)?.['Rokt.CommerceEventType'];
```
alexs-mparticle: I'm pretty certain we have an actual type for custom flags.

src/Rokt-Kit.ts
```diff
@@ -921,10 +921,29 @@ class RoktKit implements KitInterface {
...
   }
 
+  private enrichCommerceEventTypes(batch: Batch): Batch {
+    if (!batch.events) {
+      return batch;
+    }
+    for (const event of batch.events) {
+      if (event.event_type !== 'commerce_event') continue;
+
+      const data = event.data as Record<string, unknown> | undefined;
```
alexs-mparticle: I think we have a type for commerce events

src/Rokt-Kit.ts
jaissica12: good catch, added the check

src/Rokt-Kit.ts
jaissica12: Updated

src/Rokt-Kit.ts
jaissica12: Updated to use CommerceEvent from @mparticle/event-models which gives us typed custom_flags ({ [key: string]: string }) and product_action (ProductAction). The only cast remaining is as { action: string } on the assignment because ProductAction.action is typed as ProductActionActionEnum that doesn't include our new Rokt event names.

alexs-mparticle (approved):

PR #98, opened 2026-05-21
fix: use mParticle config logging flag

rmi22186:
## Summary
`isLoggingEnabled` lives on the `mParticle.config` and not in the kit settings. 

## Testing Plan
Tested locally and successfully got logs in the dashboard.

### DIFF
--- a/src/Rokt-Kit.ts
+++ b/src/Rokt-Kit.ts
@@ -31,7 +31,6 @@ interface RoktKitSettings {
   onboardingExpProvider?: string;
   loggingUrl?: string;
   errorUrl?: string;
-  isLoggingEnabled?: string | boolean;
   workspaceIdSyncApiKey?: string;
 }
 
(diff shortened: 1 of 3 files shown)


src/Rokt-Kit.ts
```diff
@@ -554,7 +553,7 @@ class ReportingTransport {
     accountId: string | null | undefined,
     rateLimiter?: RateLimiter,
   ) {
-    const isLoggingEnabled = config?.isLoggingEnabled === true || config?.isLoggingEnabled === 'true';
```
crisryantan, line 557: 🫨

crisryantan (approved):

alexs-mparticle (approved):

src/Rokt-Kit.ts
rmi22186: so even this code previously was technically wrong and no one caught it. because MPServer actually sends things back as "True" and not "true", or true.  🙃

PR #100, opened 2026-05-26
fix: prevent cached commerce attributes in Rokt placements

rmi22186:
## Summary
 - Add a built-in Rokt selectPlacements attribute persistence denylist for commerce/order attributes.
  - Prevent denylisted attributes from being cached in the kit’s userAttributes.
  - Prevent stale denylisted user attributes from being merged into later launcher.selectPlacements calls.
  - Preserve explicitly passed selectPlacements({ attributes }) values for the current call, while avoiding persistence for future calls.

## Testing Plan
* Tests added
* tested on a local app, first calling selectPlacements with code in production to load all user attributes with commerce attributes that were passed to selectplacements, then

### DIFF
--- a/src/Rokt-Kit.ts
+++ b/src/Rokt-Kit.ts
@@ -462,11 +462,11 @@ function isString(value: unknown): value is string {
   return typeof value === 'string';
 }
 
-function isSelectPlacementsAttributePersistenceDenied(key: string): boolean {
+export function isSelectPlacementsAttributePersistenceDenied(key: string): boolean {
   return SELECT_PLACEMENTS_ATTRIBUTE_PERSISTENCE_DENY_SET.has(key.toLowerCase());
 }
 
-function removeSelectPlacementsAttributePersistenceDeniedAttributes(
+export function removeSelectPlacementsAttributePersistenceDeniedAttributes(
   attributes: Record<string, unknown> | null | undefined,
 ): Record<string, unknown> {
   const filteredAttributes: Record<string, unknown> = {};
--- a/src/Rokt-Kit.ts
+++ b/src/Rokt-Kit.ts
@@ -249,6 +249,7 @@ const SELECT_PLACEMENTS_ATTRIBUTE_PERSISTENCE_DENY_LIST = [
   'cartitems',
   'ccbin',
   'confirmationref',
+  'conversiontype',
   'country',
   'couponcode',
   'currency',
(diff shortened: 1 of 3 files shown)


src/Rokt-Kit.ts
```diff
@@ -1226,6 +1272,11 @@ class RoktKit implements KitInterface {
   }
 
   public setUserAttribute(key: string, value: unknown): string {
+    if (isSelectPlacementsAttributePersistenceDenied(key)) {
```
alexs-mparticle: it looks like `isSelectPlacementsAttributePersistenceDenied` is called twice: Once here, and then again in `removeSelectPlacementsAttributePersistenceDeniedAttributes`. Wouldn't this be cleaner?

```
public setUserAttribute(key: string, value: unknown): string {
    if (!isSelectPlacementsAttributePersistenceDenied(key)) {
      this.userAttributes[key] = value;
    }
    return 'Successfully set user attribute for forwarder: ' + name;
  }
```

src/Rokt-Kit.ts
```diff
@@ -436,6 +461,27 @@ function isString(value: unknown): value is string {
...
+  for (let i = 0; i < attributeKeys.length; i++) {
+    const key = attributeKeys[i];
+    if (!isSelectPlacementsAttributePersistenceDenied(key)) {
+      filteredAttributes[key] = sourceAttributes[key];
+    }
+  }
+
+  return filteredAttributes;
+}
+
```
alexs-mparticle: We may be able to simplify this to: 

```suggestion
private filterDeniedAttributes(attrs: Record<string, unknown> | null | undefined): Record<string, unknown> {
    return Object.fromEntries(
      Object.entries(attrs || {}).filter(([key]) => !isSelectPlacementsAttributePersistenceDenied(key)),
    );
  }

```

src/Rokt-Kit.ts
```diff
@@ -1237,7 +1288,7 @@ class RoktKit implements KitInterface {
 
   private handleIdentityComplete(user: IMParticleUser, eventType: RoktIdentityEventType, callbackName: string): string {
     const filteredUser = user as FilteredUser;
-    this.userAttributes = user.getAllUserAttributes();
+    this.userAttributes = removeSelectPlacementsAttributePersistenceDeniedAttributes(user.getAllUserAttributes());
```
alexs-mparticle, line 1265: I could be wrong, but i don't see tests to cover the changes to `handleIdentityComplete`

alexs-mparticle (changes requested):

src/Rokt-Kit.ts
rmi22186: Added

src/Rokt-Kit.ts
rmi22186: seeing that Object.fromEntries is ES2019 and the package is targeting 2017, which I believe matt targeted because that's what wSDK targets

alexs-mparticle (approved):

alexs-mparticle (approved):

PR #107, opened 2026-07-31
feat: capture page views and surface as page_events in selectPlacements

alexs-mparticle:
## Summary

Records page-view events as they are logged, persists them durably to the Rokt local-session-attribute store, and surfaces them as a flat `page_events` array on the next `selectPlacements` call as targeting context.

## What's included

- **Capture** (`process`): page views (`EventDataType === 3`) are appended to a persisted `mpPageViews` list (max 25, oldest evicted), guarded so a malformed event can never throw out of the forwarder. Captures `EventName`, full `pageUrl` (`window.location.href`), `SourceMessageId`, `timestamp`, `ActiveTimeOnSite`, and `EventAttributes`.
- **Flatten** (`buildPageEvents`): each stored view is flattened into a `page_events` entry — `event_name`, `page_name` (from the `title` attribute), and `eventAttributes` exploded into `attr_`-namespaced keys so they can't clobber base fields. The raw nested `mpPageViews` store is stripped so only the flat copy is sent.
- **timeOnPage**: each entry carries a derived `timeOnPage` — the active time a page was viewed before the next page view was logged (diff of consecutive `activeTimeOnSite`). Emitted only when it's a genuine non-negative number; omitted for the still-open last entry, negative diffs, and non-numeric values. Derived at read time — nothing extra persisted.
- **Feeding selectPlacements**: relaxed `returnLocalSessionAttributes()` so the store flows through whenever it's available and populated.

## Security note (per Rokt secure-coding policy)

`pageUrl` (full URL incl. query/fragment) and `eventAttributes` are stored verbatim and may contain PII; they are persisted to browser storage and sent to Rokt. Full-URL was an explicit product choice; `sanitizeUrl()` isolates the boundary so query/fragment stripping is a one-line change later. The SDK does not run kit user-attribute filters over page-view `EventAttributes`. Design doc records this in full.

## Testing

- Navigate through a page with an MPA and then trigger`mParticle.getInstance()._Store.getLocalSessionAttributes()` in the dev console.

Expected result:

```
[
    {
        "attr_hostname": "www.domain.com",
        "event_name": "PageView",
        "page_name": "Online Store",
        "pageUrl": "https://www.domain.com/",
        "sourceMessageId": "80588459-ea2c-491a-9f41-2acc15b8cfb8",
        "timestamp": 1785526522349,
        "activeTimeOnSite": 328892,
        "timeOnPage": 35935
    },
    {
        "attr_hostname": "www.domain.com",
        "event_name": "PageView",
        "page_name": "Online Store | Some Page",
        "pageUrl": "https://www.domain.com/page_1",
        "sourceMessageId": "88a1ef32-83fd-41a8-03d8-04c4c67c879d",
        "timestamp": 1785526728989,
        "activeTimeOnSite": 364827,
        "timeOnPage": 11964
    },
    {
        "attr_hostname": "www.domain.com",
        "event_name": "PageView",
        "page_name": "Online Store | Next Page",
        "pageUrl": "https://www.domain.com/page_2",
        "sourceMessageId": "3c755531-1785-4b23-bf6c-fee3a93cc38e",
        "timestamp": 1785526922622,
        "activeTimeOnSite": 376791
    }
]

```

Verify the expected payload when making a select placements call:

```
{
  "active_time_on_site_ms": "427361",
  "page_events": "[ /* stringified array of events with timeOnPage value */ ]",
  // other placement attributes
}
```


src/Rokt-Kit.ts
```diff
@@ -1165,7 +1251,13 @@ class RoktKit implements KitInterface {
     if (!this.isKitReady()) {
       return 'Kit not ready for forwarder: ' + name;
     }
+
     if (typeof mp().Rokt?.setLocalSessionAttribute === 'function') {
+      if (event.EventDataType === MESSAGE_TYPE_PAGE_VIEW) {
+        console.warn('caputre Event', event);
```
alexs-mparticle:
```suggestion
```

src/Rokt-Kit.ts
```diff
@@ -866,12 +921,43 @@ class RoktKit implements KitInterface {
...
     }
-    if (isEmpty(this.placementEventMappingLookup) && isEmpty(this.placementEventAttributeMappingLookup)) {
-      return {};
-    }
     return mp().Rokt.getLocalSessionAttributes!();
   }
 
+  private buildPageEvents(pageViews: StoredPageView[]): Record<string, unknown>[] {
+    return pageViews.map((pv, i) => {
+      const flat: Record<string, unknown> = {};
```
alexs-mparticle: We should create an explicit type here for safety and it should include `timeOnPage` as optional.

test/src/tests.spec.ts
```diff
@@ -4939,7 +4939,7 @@ describe('Rokt Forwarder', () => {
       (window as any).mParticle.forwarder.process({
         EventName: 'Browse',
         EventCategory: EventType.Unknown,
-        EventDataType: MessageType.PageView,
+        EventDataType: MessageType.PageEvent,
```
alexs-mparticle: We shouldn't be changing these.

PR #109, opened 2026-07-31
feat: capture page views and surface as page_events in selectPlacements

alexs-mparticle:
## Summary

Captures page-view events as they are logged, persists them in the kit's **own localStorage** (separate from mParticle persistence and cookie sync), and surfaces them as a stringified `page_events` array on the next `selectPlacements` call as targeting context.

## What's included

- **Capture** (`process` → `capturePageView`): page views (`EventDataType === 3`) are appended to a kit-owned `mpPageViews` list in localStorage. Each record stores `pageUrl` (from `window.location.href`, query string stripped), `sourceMessageId`, `timestamp`, and `activeTimeOnSite`. Capture runs independently of `setLocalSessionAttribute` availability, since the kit controls its own storage.
- **Fixed count cap**: the list is capped at **25 records**, evicting the oldest first. A simple, predictable limit — no storage-backend sniffing.
- **timeOnPage** (`buildPageEvents`): each entry carries a derived `timeOnPage` — the active time a page was viewed before the next page view was logged (diff of consecutive `activeTimeOnSite`). Emitted only when it's a genuine non-negative number; omitted for the still-open last entry, negative diffs, and non-numeric values. Derived at read time — nothing extra persisted.
- **Feeding selectPlacements**: the stored page views are read back, flattened via `buildPageEvents`, and sent as a stringified `page_events` array alongside the other placement attributes.

## Storage & error handling

The kit owns its localStorage directly rather than writing through mParticle's local-session-attribute store, so page-view capture no longer touches mParticle persistence or cookie sync.

- **Read** (`readPageViewsStorage`): guarded — returns an empty array when nothing is stored, the value can't be parsed, or localStorage is unavailable (Safari private mode, storage disabled).
- **Write** (`writePageViewsStorage`): a failure (quota exceeded, storage disabled) is caught in `capturePageView` and surfaced as a `PAGE_VIEW_CAPTURE_FAILED` WARNING, bounded by the reporting service's per-severity rate limiter, without ever throwing out of the forwarder.

## Testing

Navigate through pages on an MPA, then inspect the kit-owned store in the dev console:

```js
JSON.parse(window.localStorage.getItem('mpPageViews'))
```

Expected result:

```
[
    {
        "pageUrl": "https://www.domain.com/",
        "sourceMessageId": "80588459-ea2c-491a-9f41-2acc15b8cfb8",
        "timestamp": 1785526522349,
        "activeTimeOnSite": 328892
    },
    {
        "pageUrl": "https://www.domain.com/page_1",
        "sourceMessageId": "88a1ef32-83fd-41a8-03d8-04c4c67c879d",
        "timestamp": 1785526728989,
        "activeTimeOnSite": 364827
    }
]
```

Verify the expected payload when making a selectPlacements call:

```
{
  "active_time_on_site_ms": "427361",
  "page_events": "[ /* stringified array of events, each with a derived timeOnPage */ ]",
  // other placement attributes
}
```


src/Rokt-Kit.ts
```diff
@@ -1375,11 +1481,18 @@ class RoktKit implements KitInterface {
 
     const localSessionAttributes = this.returnLocalSessionAttributes();
 
+    // Derive the flat page_events array from the stored page views, then drop the
+    // raw nested mpPageViews so Rokt receives only the flattened copy.
+    const rawPageViews = localSessionAttributes[PAGE_VIEWS_KEY];
+    const pageEvents = Array.isArray(rawPageViews) ? this.buildPageEvents(rawPageViews as StoredPageView[]) : [];
+    delete localSessionAttributes[PAGE_VIEWS_KEY];
```
alexs-mparticle: We should not be mutating this.

src/Rokt-Kit.ts
```diff
@@ -866,12 +935,44 @@ class RoktKit implements KitInterface {
...
       return {};
     }
-    if (isEmpty(this.placementEventMappingLookup) && isEmpty(this.placementEventAttributeMappingLookup)) {
-      return {};
-    }
     return mp().Rokt.getLocalSessionAttributes!();
   }
 
+  private buildPageEvents(pageViews: StoredPageView[]): PageEvent[] {
+    return pageViews.map((pv, i) => {
```
alexs-mparticle: Nit: Use pageview and index, rather than abbreviations.

src/Rokt-Kit.ts
```diff
@@ -61,6 +61,29 @@ interface RoktExtensionEntry {
   value: string;
 }
 
+// A captured page view, persisted (newest last) under PAGE_VIEWS_KEY.
+// pageUrl and eventAttributes are stored verbatim and may contain PII; they are
+// persisted to browser storage and sent to Rokt on the next selectPlacements call.
```
alexs-mparticle:
```suggestion
```

src/Rokt-Kit.ts
```diff
@@ -845,6 +880,34 @@ class RoktKit implements KitInterface {
...
+        eventAttributes: event.EventAttributes,
+      });
+
+      while (pageViews.length > MAX_PAGE_VIEWS) {
+        pageViews.shift();
+      }
+
+      mp().Rokt.setLocalSessionAttribute?.(PAGE_VIEWS_KEY, pageViews);
+    } catch (err) {
+      console.error('Rokt Kit: Failed to capture page view', err);
```
rmi22186: Instead of console.erroring, we should increase observability by sending to data dog

rmi22186: or, in addition to i should say

src/Rokt-Kit.ts
```diff
@@ -1373,13 +1468,15 @@ class RoktKit implements KitInterface {
 
     const filteredUserIdentities = this.returnUserIdentities(filteredUser);
 
-    const localSessionAttributes = this.returnLocalSessionAttributes();
+    const { [PAGE_VIEWS_KEY]: rawPageViews, ...sessionAttributes } = this.returnLocalSessionAttributes();
```
rmi22186: I was conflating `PAGE_VIEWS_KEY` and `PAGE_EVENTS_KEY`.  Maybe `LS_Page_VIEWS_KEY` is more clear to stated it's the key for the persistence?

src/Rokt-Kit.ts
```diff
@@ -866,12 +929,39 @@ class RoktKit implements KitInterface {
...
   }
 
+  private buildPageEvents(pageViews: StoredPageView[]): PageEvent[] {
+    return pageViews.map((pageView, index) => {
+      const flat: PageEvent = {
+        event_name: pageView.event_name,
+        pageUrl: pageView.pageUrl,
+        sourceMessageId: pageView.sourceMessageId,
+        timestamp: pageView.timestamp,
+        activeTimeOnSite: pageView.activeTimeOnSite,
```
rmi22186: nit: Any reason we are doing 2 transforms?  Saving to LS as a PageView, where we do some mapping, then turning a PageView into a PageEvent. Could we just save it as a PageEvent to reduce the number of schemas to keep track of?

src/Rokt-Kit.ts
```diff
@@ -845,6 +880,34 @@ class RoktKit implements KitInterface {
...
+        timestamp: event.Timestamp,
+        activeTimeOnSite: event.ActiveTimeOnSite,
+        eventAttributes: event.EventAttributes,
+      });
+
+      while (pageViews.length > MAX_PAGE_VIEWS) {
+        pageViews.shift();
+      }
+
+      mp().Rokt.setLocalSessionAttribute?.(PAGE_VIEWS_KEY, pageViews);
```
rmi22186: `setLocalSessionAttribute` only accepts primitive `AttributeValue`s (`string | number | boolean | null | undefined`) — same as every existing LSA usage (boolean flags). Passing an array of objects here works at runtime only because the setter doesn't validate, but it violates the core SDK contract.

Prefer `JSON.stringify` on write and `JSON.parse` on read so the stored value stays a string.

src/Rokt-Kit.ts
```diff
@@ -866,12 +929,39 @@ class RoktKit implements KitInterface {
...
+      };
+      if (pageView.eventAttributes) {
+        for (const [key, value] of Object.entries(pageView.eventAttributes)) {
+          if (key === PAGE_TITLE_ATTRIBUTE) {
+            continue;
+          }
+          flat[`${PAGE_EVENT_ATTR_PREFIX}${key}`] = value;
+        }
+      }
+      flat.page_name = pageView.eventAttributes?.[PAGE_TITLE_ATTRIBUTE];
```
rmi22186: you can remove the ? in `eventAttributes?` here by moving this up one line to be embedded in the `if` clause directly above

src/Rokt-Kit.ts
```diff
@@ -61,6 +61,26 @@ interface RoktExtensionEntry {
   value: string;
 }
 
+interface StoredPageView {
+  event_name: string;
+  pageUrl: string;
+  sourceMessageId: string;
+  timestamp: number;
+  activeTimeOnSite: number;
```
jamesnrokt: Do we have total time on site available too? It may be good to differentiate behaviour

src/Rokt-Kit.ts
```diff
@@ -450,6 +479,12 @@ function isString(value: unknown): value is string {
   return typeof value === 'string';
 }
 
+// Isolates page-view URL handling. Returns the URL verbatim for now; tightening
+// to strip query/fragment (which may carry PII) later is a one-line change here.
```
jamesnrokt: I don't think this change can wait

src/Rokt-Kit.ts
```diff
@@ -61,6 +61,26 @@ interface RoktExtensionEntry {
...
+  event_name: string;
+  pageUrl: string;
+  sourceMessageId: string;
+  timestamp: number;
+  activeTimeOnSite: number;
+  eventAttributes?: { [key: string]: string };
+}
+
+interface PageEvent {
+  event_name: string;
```
jamesnrokt: Won't this always be the same as they're all going to be page views? If so do we want this repetition?

src/Rokt-Kit.ts
```diff
@@ -61,6 +61,26 @@ interface RoktExtensionEntry {
   value: string;
 }
 
+interface StoredPageView {
+  event_name: string;
+  pageUrl: string;
+  sourceMessageId: string;
+  timestamp: number;
+  activeTimeOnSite: number;
+  eventAttributes?: { [key: string]: string };
```
jamesnrokt: In the case of auto page views will there be anything useful in this?

src/Rokt-Kit.ts
```diff
@@ -244,6 +264,15 @@ const ROKT_INTEGRATION_SCRIPT_ID = 'rokt-launcher';
 const ROKT_THANK_YOU_ELEMENT_SCRIPT_ID = 'rokt-thank-you-element';
 const USER_IDENTIFIED_IN_WORKSPACE_KEY = 'userIdentifiedInWorkspace';
 
+const MESSAGE_TYPE_PAGE_VIEW = 3; // mParticle MessageType.PageView
+const PAGE_VIEWS_KEY = 'mpPageViews';
+const MAX_PAGE_VIEWS = 25;
```
jamesnrokt: Probably needs a comment on how / why this max is set to 25

src/Rokt-Kit.ts
```diff
@@ -866,12 +929,39 @@ class RoktKit implements KitInterface {
...
+          if (key === PAGE_TITLE_ATTRIBUTE) {
+            continue;
+          }
+          flat[`${PAGE_EVENT_ATTR_PREFIX}${key}`] = value;
+        }
+      }
+      flat.page_name = pageView.eventAttributes?.[PAGE_TITLE_ATTRIBUTE];
+
+      const next = pageViews[index + 1];
+      if (next && typeof next.activeTimeOnSite === 'number' && typeof pageView.activeTimeOnSite === 'number') {
```
jamesnrokt: `StoredPageView.activeTimeOnSite` is typed to a non-optional number do we need the guard? If we do then we should re-evaluate the type declaration

src/Rokt-Kit.ts
```diff
@@ -1373,13 +1468,15 @@ class RoktKit implements KitInterface {
...
+    const { [PAGE_VIEWS_KEY]: rawPageViews, ...sessionAttributes } = this.returnLocalSessionAttributes();
+    const pageEvents = Array.isArray(rawPageViews) ? this.buildPageEvents(rawPageViews as StoredPageView[]) : [];
 
     const selectPlacementsAttributes: Record<string, unknown> = {
       ...(filteredUserIdentities as Record<string, unknown>),
       ...filteredAttributes,
       ...optimizelyAttributes,
-      ...localSessionAttributes,
+      ...sessionAttributes,
+      ...(pageEvents.length ? { [PAGE_EVENTS_KEY]: pageEvents } : {}),
```
jamesnrokt: The PR description suggests this will be stringified is this correct?

src/Rokt-Kit.ts
```diff
@@ -866,12 +929,39 @@ class RoktKit implements KitInterface {
     if (!mp().Rokt || typeof mp().Rokt.getLocalSessionAttributes !== 'function') {
       return {};
     }
-    if (isEmpty(this.placementEventMappingLookup) && isEmpty(this.placementEventAttributeMappingLookup)) {
-      return {};
-    }
     return mp().Rokt.getLocalSessionAttributes!();
   }
 
+  private buildPageEvents(pageViews: StoredPageView[]): PageEvent[] {
```
jamesnrokt: Might be worth breaking this function up. It does 3 distinct pieces of functionality

src/Rokt-Kit.ts
```diff
@@ -866,12 +929,39 @@ class RoktKit implements KitInterface {
...
+    return pageViews.map((pageView, index) => {
+      const flat: PageEvent = {
```
jamesnrokt: nit: `flat` doesn't really describe what the variable is could it just be called `pageEvent`?

src/Rokt-Kit.ts
```diff
@@ -866,12 +929,39 @@ class RoktKit implements KitInterface {
     if (!mp().Rokt || typeof mp().Rokt.getLocalSessionAttributes !== 'function') {
       return {};
     }
-    if (isEmpty(this.placementEventMappingLookup) && isEmpty(this.placementEventAttributeMappingLookup)) {
```
jamesnrokt, line 869: Any reason for this removal? Won't this cause every local session attribute to flow into select placements for all customers?

src/Rokt-Kit.ts
```diff
@@ -845,6 +880,34 @@ class RoktKit implements KitInterface {
...
   }
 
+  // Appends a page-view record to the persisted list under PAGE_VIEWS_KEY,
+  // capping at MAX_PAGE_VIEWS (oldest evicted). Wrapped so a malformed event
+  // can never throw out of the forwarder. Callers must confirm the event is a
+  // page view and that setLocalSessionAttribute is available.
+  private capturePageView(event: SDKEvent): void {
+    try {
+      const existing = mp().Rokt.getLocalSessionAttributes?.()?.[PAGE_VIEWS_KEY];
+      const pageViews: StoredPageView[] = Array.isArray(existing) ? (existing as StoredPageView[]) : [];
```
jamesnrokt: Shouldn't we be operating on a copy of existing rather than directly?

jamesnrokt (changes requested):
Requesting changes as we definitely need the cleansing of the URL in before go live

src/Rokt-Kit.ts
alexs-mparticle: It's no longer necessary since the intention of this guard was to only call when we were checking for specific mappings in LSA. Now that we're storing page view events, it's adding an unnecessary guard. Maybe a better guard would be to check if LSA is empty?

src/Rokt-Kit.ts
```diff
@@ -845,6 +883,56 @@ class RoktKit implements KitInterface {
...
+        eventAttributes: event.EventAttributes,
+      });
+
+      while (pageViews.length > MAX_PAGE_VIEWS) {
+        pageViews.shift();
+      }
+
+      mp().Rokt.setLocalSessionAttribute?.(LS_PAGE_VIEWS_KEY, JSON.stringify(pageViews));
+    } catch (err) {
+      this.errorReportingService?.report({
```
alexs-mparticle, line 932: We should capture the page url if we can.

src/Rokt-Kit.ts
alexs-mparticle: Let's discuss

src/Rokt-Kit.ts
alexs-mparticle: Probably not for auto page views, but this logic needs to support manual page views as well, as we don't have a way to differentiate.

jamesnrokt: I think we're good without this given auto page views will solve most of the data problems long term anyway. Additionally this saves storage on the cookie sync side too

src/Rokt-Kit.ts
alexs-mparticle: Since manual page views are also coming in, this could be overridden by a customer

src/Rokt-Kit.ts
alexs-mparticle: Since we're parsing/stringifying this every time we read/write to local storage, there's a risk it could get mangled or become `NaN`.

src/Rokt-Kit.ts
alexs-mparticle: Discussed and decided to mirror the Forwarder by removing query params.

src/Rokt-Kit.ts
alexs-mparticle: Correct. I may be misremembering but I thought there was a separate process that stringifies objects in the attributes. I'll make sure it's explicit here to avoid regressions.

src/Rokt-Kit.ts
alexs-mparticle: `ActiveTimeOnSite` is essentially the Total Time on Site as it's calculated in the core SDK.

https://github.com/mParticle/mparticle-web-sdk/blob/v2.75.0/src/foregroundTimeTracker.ts#L105

https://github.com/mParticle/mparticle-web-sdk/blob/v2.75.0/src/serverModel.ts#L313

src/Rokt-Kit.ts
alexs-mparticle: Now that we've removed the manual page view fields, we can reduce the scope of this function.

src/Rokt-Kit.ts
jamesnrokt: Understood and caught up in person. This would need a separate PR to capture the users full duration on site including background time as this would be beneficial for understanding user behavior.

While we can do this in this PR by taking the current timestamp and deducting the timestamp of the first page view we will tackle the ask separately.

src/Rokt-Kit.ts
jamesnrokt: Unresolved as while the description change makes sense we likely need real figures to make sure we're not overloading storage or cookie sync

src/Rokt-Kit.ts
jamesnrokt: Looks like this was fully removed. Resolving

src/Rokt-Kit.ts
jamesnrokt: Noted it's removed. Resolving

src/Rokt-Kit.ts
jamesnrokt: Now correctly named. Resolving

src/Rokt-Kit.ts
jamesnrokt: Seeing stringify now

src/Rokt-Kit.ts
jamesnrokt: Resolving as you changed implementation

src/Rokt-Kit.ts
jamesnrokt: Resolving as it's captured as a separate task

src/Rokt-Kit.ts
```diff
@@ -1162,9 +1267,12 @@ class RoktKit implements KitInterface {
   }
 
   public process(event: SDKEvent): string {
```
jamesnrokt, line 1301: With the isKitReady change the block below now also runs pre-ready, and `process()` always returns `'Successfully sent to forwarder'`, so the core SDK loses the not-ready signal entirely.

src/Rokt-Kit.ts
```diff
@@ -61,6 +61,16 @@ interface RoktExtensionEntry {
   value: string;
 }
 
+interface PageEvent {
+  pageUrl: string;
+  sourceMessageId: string;
+  timestamp: number;
+  activeTimeOnSite: number;
```
jamesnrokt: Coming back to this with your NaN point resolved on another comment, which I agree with as that's exactly why the current state looks inconsistent to me.

Can we mark it `activeTimeOnSite?: number` so the type matches what the storage round-trip can actually produce?

jamesnrokt (commented):
Final remaining comment I know you're already addressing is clearing out storage on session

src/Rokt-Kit.ts
alexs-mparticle: I think I found a happy middle to deal with this case using an [`isFinite`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite) check.

src/Rokt-Kit.ts
alexs-mparticle: Restored

src/Rokt-Kit.ts
jamesnrokt: The storage reasoning makes sense — the test comment on `JSON.stringify(NaN) === 'null'` is a good catch and coercing does keep the stored type honest.

The problem is what `0` does downstream in `buildPageEvents`. If a record's `ActiveTimeOnSite` is missing it stores `0`, and the next record's diff is then measured against that `0`:

```
A: ActiveTimeOnSite undefined → stored 0
B: ActiveTimeOnSite 300000    → stored 300000
→ A.timeOnPage = 300000 - 0 = 300000
```

So page A reports a 5 minute dwell time that never happened. Before the coercion this came out as `NaN`, failed the `diff >= 0` check and `timeOnPage` was correctly omitted — we lost the value but didn't invent one. `activeTimeOnSite: 0` on the wire has the same issue: it's indistinguishable from a genuine zero.

The new test only asserts the single-record write, so this isn't covered.

I think optional is still the better trade here — omit when non-finite and guard the diff on both sides being numbers. That keeps only finite numbers in storage, same as now, but keeps "unknown" distinguishable from "zero" instead of fabricating a duration. Can we go back to `activeTimeOnSite?: number`?

src/Rokt-Kit.ts
```diff
@@ -1162,9 +1272,31 @@ class RoktKit implements KitInterface {
   }
 
   public process(event: SDKEvent): string {
+    if (!this.isTargetingDisabled()) {
```
jamesnrokt, line 1302: Would be worth either adding a clear of the store if it's disabled or adding the same gate for when we're about to send the events

alexs-mparticle: From a privacy standpoint, I agree, but I think it would be more performant to clear out storage on init rather than every time an event or placement is triggered. I'll add it to the kit init flow so we're not spamming storage..

src/Rokt-Kit.ts
alexs-mparticle: Good call out. I restored this and cleaned up the checks.

rmi22186 (approved):
I noticed that Event.EventAttributes were removed between commits at some point. If that is the intent, then this LGTM

jamesnrokt (approved):
Thanks for your efforts on this - apologies for the many questions

PR #112, opened 2026-08-07
feat: capture page title and canonical URL in page view events

alexs-mparticle:
## Summary

Extends the auto page-view capture feature (added in #109) to also record the **page title** and **canonical URL** on each page-view event.

Both fields are nested inside each `page_events` record — alongside the existing `pageUrl`, `sourceMessageId`, `timestamp`, etc. — and surface in the `page_events` attribute sent to `selectPlacements`. No new top-level attributes are introduced.

## Changes

- Add optional `pageTitle` and `canonicalUrl` to the `PageEvent` interface.
- `capturePageView()` reads `document.title` and `<link rel="canonical">`.
- New `readCanonicalUrl()` helper resolves the canonical href to absolute and sanitizes it through the **same** `sanitizeUrl()` used for `pageUrl` (query string stripped, hash fragment retained) — keeping canonical handling consistent with `pageUrl`.
- `buildPageEvents()` carries both fields through to the transmitted payload.
- Both fields are omitted when unavailable, matching how `activeTimeOnSite` is handled.

## Example `page_events` record

```json
{
  "pageUrl": "https://example.com/checkout#section",
  "sourceMessageId": "...",
  "timestamp": 1712345678000,
  "pageTitle": "Checkout",
  "canonicalUrl": "https://example.com/canonical#section",
  "activeTimeOnSite": 4200
}
```

## Tests

Three new tests: captures title + canonical, omits both when absent, and carries them through `selectPlacements`. Full suite: 231/231 passing. Lint and build clean.

## Note on sanitization

`canonicalUrl` reuses `sanitizeUrl()` for consistency with `pageUrl`, which strips the query string but **retains** the hash fragment (per the existing `pageUrl` behavior/tests). If we'd prefer to strip the fragment from canonical too, that's a small follow-up but would diverge from `pageUrl` handling.

### DIFF
--- a/src/Rokt-Kit.ts
+++ b/src/Rokt-Kit.ts
@@ -26,13 +26,15 @@ import {
 
 import {
   PageEvent,
+  buildPageEvents,
   migrateLegacyPageViewStorage,
   loadPageViews,
   writePageViews,
   clearPageViews,
+  readCanonicalUrl,
 } from './pageViewStorage';
 
-import { isObject, isString, isEmpty } from './utils';
+import { isObject, isString, isEmpty, sanitizeUrl } from './utils';
 
 interface RoktKitSettings {
   accountId: string;
@@ -445,28 +447,6 @@ function hashEventMessage(messageType: number, eventType: number, eventName: str
   return mp().generateHash([messageType, eventType, eventName].join(''));
 }
 
-// Strips the query string from a page-view URL before it is persisted and sent
-// to Rokt, since query params commonly carry PII (emails, tokens, order refs).
-// Returns the input unchanged if it can't be parsed as a URL.
-function sanitizeUrl(href: string): string {
-  try {
-    const url = new URL(href);
-    url.search = '';
-    return url.toString();
-  } catch {
-    return href;
-  }
-}
-
-function readCanonicalUrl(): string | undefined {
-  const link = document.querySelector<HTMLLinkElement>('link[rel="canonical"]');
-  const href = link?.href;
-  if (!href) {
-    return undefined;
-  }
-  return sanitizeUrl(href);
-}
-
 function generateIntegrationName(customIntegrationName?: string): string {
   const coreSdkVersion = mp().getVersion();
   const kitVersion = process.env.PACKAGE_VERSION;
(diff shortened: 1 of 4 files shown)


src/Rokt-Kit.ts
```diff
@@ -917,6 +928,16 @@ class RoktKit implements KitInterface {
         timestamp: event.Timestamp,
       };
 
+      const pageTitle = document.title;
```
rmi22186: nit: Should this come from event since it shoudl be there already in the edge case small chance it's different from document.title?

```suggestion
      const pageTitle = event.EventAttributes?.title || document.title;
```

rmi22186 (changes requested):
Adding the `pageTitle` and `canonicalUrl` will only make that error we were seeing for quota limit in setting local storage worse. I wouldn't merge this without including the fix for resolving reducing the rate we see the error.

src/pageViewStorage.ts
```diff
@@ -57,3 +60,35 @@ export function writePageViews(pageViews: PageEvent[]): boolean {
...
+    const diff = hasActiveTime && hasNextActiveTimeOnSite ? nextActiveTimeOnSite - activeTimeOnSite : undefined;
+
+    return {
+      pageUrl: pageView.pageUrl,
+      sourceMessageId: pageView.sourceMessageId,
+      timestamp: pageView.timestamp,
+      ...(pageView.pageTitle !== undefined ? { pageTitle: pageView.pageTitle } : {}),
+      ...(pageView.canonicalUrl !== undefined ? { canonicalUrl: pageView.canonicalUrl } : {}),
+      ...(hasActiveTime ? { activeTimeOnSite } : {}),
+      ...(diff !== undefined && diff >= 0 ? { activeTimeOnPage: diff } : {}),
```
rmi22186, line 82: this activeTimeOnPage - am I right that this is technically activeTimeOnThePreviousPage?

rmi22186 (approved):

src/pageViewStorage.ts
alexs-mparticle: technically...

PR #113, opened 2026-08-11
feat: namespace page-view localStorage key and migrate legacy key

alexs-mparticle:
## What

Moves the kit's persisted page-view history out of the unprefixed, top-level `mpPageViews` localStorage key into a single kit-owned namespace object stored under **`mp-rokt-kit`**, with page views held in its **`pageViews`** field. This establishes one `mp-rokt-kit` object where future kit-owned state can live as sibling fields, and adds a one-time migration so existing history survives the move.

## Changes

- **Namespace object:** page views are now stored as `mp-rokt-kit` → `{ "pageViews": [...] }`. Future kit-owned state becomes additional fields on the same object rather than new top-level keys.
- **Generalized storage accessors (`src/storage.ts`):** key-agnostic JSON read/write/remove helpers, plus namespaced-field helpers that read-modify-write a single field while preserving its siblings. Page-view semantics stay in the callers.
- **One-time legacy migration:** existing `mpPageViews` history is adopted into the new field on first read, and the legacy key is swept. The migration is a self-contained shim marked for removal after 2027-02-11.
- **Best-effort persistence:** page-view storage failures no longer surface as reported errors — selection proceeds with page events omitted.

## Migration semantics

The stored field value is a bare array of page views; the only new structure is the enclosing `mp-rokt-kit` object.

| Legacy `mpPageViews` | New `mp-rokt-kit.pageViews` | Result |
|---|---|---|
| present | absent | new field adopts legacy value; legacy key swept |
| present | present | new wins; legacy key swept (discarded) |
| absent | either | no-op |

Note: with targeting disabled, an orphaned legacy `mpPageViews` may remain until the next read or `SessionEnd` sweeps it — benign.

## Tests

Added a legacy-migration suite (adopt, new-wins, no-op, `SessionEnd` sweep) and `src/storage.spec.ts` covering the storage helpers (round-trip, malformed JSON, access/quota errors, sibling preservation, namespace cleanup). Existing page-view tests repointed to the new key/field shape.

**Verify:** lint · build · tests all pass.

### DIFF
--- a/src/Rokt-Kit.ts
+++ b/src/Rokt-Kit.ts
@@ -330,15 +330,17 @@ function readJSON(key: string): unknown {
   }
 }
 
-// writeJSON/removeKey swallow storage failures (private mode, quota, access
-// denied) rather than throw: persisted page views are a best-effort cache, and
-// a failed write/remove must never break the caller. Failures are intentionally
-// not reported here — revisit if these keys ever hold must-persist data.
-function writeJSON(key: string, value: unknown): void {
+// writeJSON/removeKey never throw on storage failure (private mode, quota,
+// access denied): persisted page views are a best-effort cache, and a failed
+// write/remove must not break the caller. writeJSON returns whether the write
+// landed so callers can surface a diagnostic; removeKey is fire-and-forget (a
+// failed remove only risks orphaned data, resolved by the next write/clear).
+function writeJSON(key: string, value: unknown): boolean {
   try {
     window.localStorage.setItem(key, JSON.stringify(value));
+    return true;
   } catch {
-    // no-op
+    return false;
   }
 }
 
@@ -378,8 +380,8 @@ function loadPageViews(): PageEvent[] {
   return Array.isArray(parsed) ? (parsed as PageEvent[]) : [];
 }
 
-function writePageViewsStorage(pageViews: PageEvent[]): void {
-  writeJSON(LS_PAGE_VIEWS_KEY, pageViews);
+function writePageViewsStorage(pageViews: PageEvent[]): boolean {
+  return writeJSON(LS_PAGE_VIEWS_KEY, pageViews);
 }
 
 function clearPageViewsStorage(): void {
@@ -981,13 +983,22 @@ class RoktKit implements KitInterface {
         pageViews.shift();
       }
 
-      writePageViewsStorage(pageViews);
+      if (!writePageViewsStorage(pageViews)) {
+        // Best-effort cache: a failed persist only means fewer page events are
+        // forwarded later. Surface it as a diagnostic INFO log, not an error.
+        this.loggingService?.log({
+          message: `Rokt Kit: Failed to persist page view for ${pageUrl}`,
+          code: 'PAGE_VIEW_CAPTURE_FAILED',
+        });
+      }
     } catch (err) {
-      this.errorReportingService?.report({
-        message: `Rokt Kit: Failed to capture page view for ${pageUrl}`,
+      // sanitizeUrl / loadPageViews (legacy migration) failure — same best-effort
+      // posture: capture is skipped, no user-facing breakage. Diagnostic INFO log.
+      this.loggingService?.log({
+        message: `Rokt Kit: Failed to capture page view for ${pageUrl}: ${
+          err instanceof Error ? err.message : String(err)
+        }`,
         code: 'PAGE_VIEW_CAPTURE_FAILED',
-        severity: WSDKErrorSeverity.INFO,
-        stackTrace: err instanceof Error ? err.stack : undefined,
       });
     }
   }
@@ -1596,11 +1607,14 @@ class RoktKit implements KitInterface {
     try {
       storedPageViews = loadPageViews();
     } catch (err) {
-      this.errorReportingService?.report({
-        message: 'Rokt Kit: Failed to load page views for selectPlacements',
+      // A read/migration failure is a benign, best-effort miss (fall back to no
+      // page events) — not an SDK error. Surface it as a diagnostic INFO log
+      // rather than an error report.
+      this.loggingService?.log({
+        message: `Rokt Kit: Failed to load page views for selectPlacements: ${
+          err instanceof Error ? err.message : String(err)
+        }`,
         code: 'PAGE_VIEW_CAPTURE_FAILED',
-        severity: WSDKErrorSeverity.INFO,
-        stackTrace: err instanceof Error ? err.stack : undefined,
       });
     }
     const pageEvents = this.buildPageEvents(storedPageViews);
--- a/src/Rokt-Kit.ts
+++ b/src/Rokt-Kit.ts
@@ -322,11 +322,11 @@ function migrateLegacyPageViewStorage(loggingService: LoggingService | null): vo
   }
 
   const alreadyMigrated = readNamespacedField(LS_NAMESPACE_KEY, LS_PAGE_VIEWS_FIELD) !== undefined;
-  const needsAdoption = !alreadyMigrated && Array.isArray(legacyViews);
+  const needsMigration = !alreadyMigrated && Array.isArray(legacyViews);
 
-  if (needsAdoption) {
-    const adopted = writeNamespacedField(LS_NAMESPACE_KEY, LS_PAGE_VIEWS_FIELD, legacyViews);
-    if (!adopted) {
+  if (needsMigration) {
+    const migrated = writeNamespacedField(LS_NAMESPACE_KEY, LS_PAGE_VIEWS_FIELD, legacyViews);
+    if (!migrated) {
       loggingService?.log({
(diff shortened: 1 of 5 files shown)


src/Rokt-Kit.ts
```diff
@@ -1533,7 +1588,22 @@ class RoktKit implements KitInterface {
...
-    const pageEvents = this.buildPageEvents(readPageViewsStorage());
+    // loadPageViews() runs the legacy migration, which touches localStorage
+    // directly and can throw (e.g. access denied, quota). A read failure must
+    // not break placement selection — fall back to no page events, matching the
+    // best-effort posture of the capture and clear paths.
+    let storedPageViews: PageEvent[] = [];
+    try {
+      storedPageViews = loadPageViews();
+    } catch (err) {
+      this.errorReportingService?.report({
```
alexs-mparticle: Note to self: This should be a diagnostic info log

src/Rokt-Kit.ts
```diff
@@ -257,14 +259,9 @@ const USER_IDENTIFIED_IN_WORKSPACE_KEY = 'userIdentifiedInWorkspace';
...
 const MESSAGE_TYPE_SESSION_END = 2; // mParticle MessageType.SessionEnd
-// localStorage key under which captured page views are persisted (as a JSON
-// string). The kit owns this storage directly — separate from mParticle's
-// cookie/localStorage — so page-view capture does not affect mParticle
-// persistence or cookie sync. Distinct from PAGE_EVENTS_KEY, which is the
-// flattened wire shape sent to Rokt on selectPlacements.
-const LS_PAGE_VIEWS_KEY = 'mpPageViews';
-// Fixed cap on the number of persisted page views (oldest evicted first). Code
-// constant, not a kit setting — change it here.
+const LS_PAGE_VIEWS_KEY = 'mp-rokt-kit.pageViews';
```
crisryantan: do we care about the switch from `mpPageViews` to `mp-rokt-kit.pageViews`? just thinking of old instances that referenced this 🤔

src/Rokt-Kit.ts
```diff
@@ -311,25 +308,34 @@ function mp(): MParticleExtended {
...
-  } catch {
-    return [];
+// TODO: remove after 2027-02-11 — one-time migration of the legacy 'mpPageViews'
+// key to the prefixed LS_PAGE_VIEWS_KEY. Everything migration-related is confined
+// to this function + LEGACY_PAGE_VIEWS_KEY so it can be deleted as a single unit.
+// Unconditional (no freshness gate): staleness is mParticle's job — a timed-out
+// prior session fires SessionEnd (→ clear) before selectPlacements runs.
+function migrateLegacyPageViewStorage(): void {
+  const legacy = window.localStorage.getItem(LEGACY_PAGE_VIEWS_KEY);
+  if (legacy === null) {
```
crisryantan: nit, we can just do `if (!legacy)`

src/Rokt-Kit.ts
alexs-mparticle: That's why we have the migration below.

src/storage.ts
```diff
@@ -0,0 +1,25 @@
+export function readJSON(key: string): unknown {
+  try {
+    const stored = window.localStorage.getItem(key);
+    return stored === null ? null : JSON.parse(stored);
+  } catch {
+    return null;
+  }
+}
+
+export function writeJSON(key: string, value: unknown): boolean {
```
crisryantan, line 10: Curious on returning `boolean` here as this seems more of a `void` and we can just do a `readJSON` downstream if it exists

alexs-mparticle: I went back on forth on the implementation and just landed on this for simplicity. It may be a more relevant change when we rebase for #114.

crisryantan (approved):

jaissica12 (approved):
lgtm! reminder: can we also make same PR on mP core sdk to align the kit for V3 release

src/Rokt-Kit.ts
```diff
@@ -309,33 +310,39 @@ function mp(): MParticleExtended {
...
 // prior session fires SessionEnd (→ clear) before selectPlacements runs.
 function migrateLegacyPageViewStorage(): void {
   const legacy = window.localStorage.getItem(LEGACY_PAGE_VIEWS_KEY);
   if (!legacy) {
     return;
   }
-  if (window.localStorage.getItem(LS_PAGE_VIEWS_KEY) === null) {
-    window.localStorage.setItem(LS_PAGE_VIEWS_KEY, legacy);
+  if (readNamespacedField(LS_NAMESPACE_KEY, LS_PAGE_VIEWS_FIELD) === undefined) {
+    const legacyViews = readJSON(LEGACY_PAGE_VIEWS_KEY);
```
rmi22186: This seems duplicative of line 319. I recommend changing line 319 to be 324, then you don't need line 324 anymore.

src/Rokt-Kit.ts
```diff
@@ -309,33 +310,39 @@ function mp(): MParticleExtended {
...
+    // A failed adopt must not sweep the legacy key — throw so the caller's
```
rmi22186: Can you explain what this comment means?

src/Rokt-Kit.ts
```diff
@@ -309,33 +310,39 @@ function mp(): MParticleExtended {
...
+    // diagnostic path handles it and the next read can retry.
+    if (Array.isArray(legacyViews) && !writeNamespacedField(LS_NAMESPACE_KEY, LS_PAGE_VIEWS_FIELD, legacyViews)) {
+      throw new Error('Rokt Kit: Failed to migrate legacy page-view storage');
```
rmi22186: Should we also log the error to DD?

test/src/tests.spec.ts
```diff
@@ -5639,8 +5639,10 @@ describe('Rokt Forwarder', () => {
 
     describe('page view capture', () => {
       const readStoredPageViews = () => {
-        const raw = window.localStorage.getItem('mpPageViews');
-        return raw === null ? null : JSON.parse(raw);
+        const raw = window.localStorage.getItem('mp-rokt-kit');
+        if (raw === null) return null;
+        const parsed = JSON.parse(raw);
+        return parsed && typeof parsed === 'object' && 'pageViews' in parsed ? parsed.pageViews : null;
```
rmi22186: Looking at this again, seems like you shouldn't recreate this method but instead import from src/storage instead using `readJSON` to keep things more DRY

test/src/tests.spec.ts
```diff
@@ -5815,6 +5817,172 @@ describe('Rokt Forwarder', () => {
...
+          return raw === null ? null : JSON.parse(raw);
+        };
+
+        const readNamespacedPageViews = () => {
+          const blob = readRaw(NS_KEY);
+          return blob && typeof blob === 'object' && 'pageViews' in blob ? blob.pageViews : null;
+        };
+
+        const seedNamespacedPageViews = (views: unknown) =>
+          window.localStorage.setItem(NS_KEY, JSON.stringify({ pageViews: views }));
```
rmi22186: Similar to comment above, why not use the existing methods in storage.ts?

test/src/tests.spec.ts
```diff
@@ -5815,6 +5817,172 @@ describe('Rokt Forwarder', () => {
...
+
+        it('adopts legacy history into the new key and sweeps the legacy key on read', async () => {
+          const seeded = [
+            {
+              pageUrl: 'https://example.com/legacy',
+              sourceMessageId: 'legacy-1',
+              timestamp: 1712345678000,
+            },
+          ];
+          window.localStorage.setItem(LEGACY_KEY, JSON.stringify(seeded));
```
rmi22186: Use the help you made - `seedNamespacedPageViews`

src/Rokt-Kit.ts
```diff
@@ -316,17 +316,16 @@ function mp(): MParticleExtended {
...
-    const legacyViews = readJSON(LEGACY_PAGE_VIEWS_KEY);
-    // A failed adopt must not sweep the legacy key — throw so the caller's
-    // diagnostic path handles it and the next read can retry.
-    if (Array.isArray(legacyViews) && !writeNamespacedField(LS_NAMESPACE_KEY, LS_PAGE_VIEWS_FIELD, legacyViews)) {
-      throw new Error('Rokt Kit: Failed to migrate legacy page-view storage');
-    }
+  if (
+    readNamespacedField(LS_NAMESPACE_KEY, LS_PAGE_VIEWS_FIELD) === undefined &&
+    Array.isArray(legacyViews) &&
+    !writeNamespacedField(LS_NAMESPACE_KEY, LS_PAGE_VIEWS_FIELD, legacyViews)
```
rmi22186: Maybe it's personal opinion, but having 2 truthy statements, then a `writeNamespacedField` which does 2 things - write to LS, and return true, makes this hard to read.  That manes nesting the if for !writeNamespacedField, but I think from a readability standpoint it's a win.

rmi22186 (approved):

PR #114, opened 2026-08-11
feat: cap page-view storage by byte budget instead of fixed count

alexs-mparticle:
## What & why

Persisted page-view history was capped at a fixed 25 records. This replaces that cap with a **byte budget** (~100 KB): we keep more history when there's room to spare — which improves targeting — while staying a considerate tenant in the customer's shared-origin `localStorage` instead of assuming the space is ours.

## Behaviour

- Oldest page views are evicted first to keep our footprint under the budget.
- If the browser reports its own storage quota is full, we drop oldest and retry so the write never fails outright — always keeping at least the most recent page view.
- If even the newest page view can't be stored, we log at INFO and carry on. Nothing is ever thrown into the host page.

The `selectPlacements` send path is unchanged — it's now naturally bounded by what's in storage.

## Tests

Byte-budget eviction (oldest-first, stays under budget), newest always retained, quota-exceeded → evict-and-retry, and the give-up path logs at INFO without throwing. Full suite, lint, and build pass.

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -1,3 +1,5 @@
+import { isObject } from './utils';
+
 export function readJSON(key: string): unknown {
   try {
     const stored = window.localStorage.getItem(key);
@@ -24,25 +26,21 @@ export function removeKey(key: string): void {
   }
 }
 
-function isPlainObject(value: unknown): value is Record<string, unknown> {
-  return typeof value === 'object' && value !== null && !Array.isArray(value);
-}
-
 export function readNamespacedField(namespaceKey: string, field: string): unknown {
   const blob = readJSON(namespaceKey);
-  return isPlainObject(blob) ? blob[field] : undefined;
+  return isObject(blob) ? blob[field] : undefined;
 }
 
 export function writeNamespacedField(namespaceKey: string, field: string, value: unknown): boolean {
   const blob = readJSON(namespaceKey);
-  const next = isPlainObject(blob) ? { ...blob } : {};
+  const next = isObject(blob) ? { ...blob } : {};
   next[field] = value;
   return writeJSON(namespaceKey, next);
 }
 
 export function removeNamespacedField(namespaceKey: string, field: string): void {
   const blob = readJSON(namespaceKey);
-  if (!isPlainObject(blob) || !(field in blob)) {
+  if (!isObject(blob) || !(field in blob)) {
     return;
   }
   const next = { ...blob };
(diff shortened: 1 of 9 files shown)


test/src/pageViewStorage.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+import { readJSON, writeNamespacedField } from '../../src/storage';
+import {
+  PageViewLogger,
+  migrateLegacyPageViewStorage,
+  loadPageViews,
+  writePageViews,
+  clearPageViews,
+} from '../../src/pageViewStorage';
+
+const NS = 'mp-rokt-kit';
```
alexs-mparticle: Note to self: No short names.

test/src/pageViewStorage.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+
+      const stored = loadPageViews(null);
+      expect(JSON.stringify(stored).length).toBeLessThanOrEqual(100 * 1024);
+      expect(stored.length).toBeLessThan(30);
+      expect(stored[stored.length - 1].sourceMessageId).toBe('seed-29');
+    });
+  });
+
+  describe('clearPageViews', () => {
+    it('removes the page-view field', () => {
```
rmi22186, line 106: nit - this doesn't prove that the page_view has cleared, but rather there is nothing in the page_view. if you are removing it, you'd get the localStorage key, then check it doesn't have the key on it

src/storage.ts
```diff
@@ -53,3 +51,29 @@ export function removeNamespacedField(namespaceKey: string, field: string): void
...
+  let overBudget = JSON.stringify(records).length > maxBytes;
+  while (overBudget && evictOldest()) {
+    overBudget = JSON.stringify(records).length > maxBytes;
+  }
+
+  let written = writeNamespacedField(namespaceKey, field, records);
+  while (!written && evictOldest()) {
+    written = writeNamespacedField(namespaceKey, field, records);
+  }
+  return written;
```
rmi22186, line 84: so the idea here is - evict the oldest while we are over budget, and then, once we are under budget, if we fail to write to local storage, remove each item until we are under budget again based on what space is left?

alexs-mparticle: Correct, although I think I can probably clean this up a bit to make it more clear and concise.

rmi22186 (approved):

crisryantan (approved):

PR #116, opened 2026-08-17
feat: send the current mParticle session id as a selectPlacements attribute

jamesnrokt:
## Summary

The kit read the mParticle session id once, when creating the launcher, and passed it through as the `mpSessionId` launcher option. That option is no longer used, and reading the value at launcher-creation time meant it went stale for the life of the page: mParticle mints a new session id on inactivity timeout (`sessionTimeout`, 30 minutes by default), when another tab starts a new session, and after an explicit `endSession()` / `startNewSession()`. None of that reached the launcher.

This PR:

- Sends the session id as the **`mparticle_session_id` attribute** on each `selectPlacements` call. Because it is read per call, a rotated session is reflected on the next call.
- Removes the `mpSessionId` launcher option from `attachLauncher`.
- Adds `readMpSessionId()`, which calls `SessionManager.getSessionId()`.

Review follow-ups applied:

- Dropped the `getSession()` fallback. The kit requires `@mparticle/web-sdk` `^2.62.0` and `getSessionId()` has shipped since `v2.23.3`, so the fallback was unreachable for every supported core version — and this removes the kit's last caller of a deprecated method.
- Added `isFunction` to `src/utils.ts` alongside `isObject` / `isString` / `isEmpty`, and used it for all five `typeof x === 'function'` checks in the kit rather than leaving the new one inconsistent with the rest.

## Testing Plan

- `npm test` — 297/297 pass. `npm run lint` and `npm run build` clean.
- New unit tests cover: the session id current at the time of each call (asserted across a rotation between two calls), the attribute being omitted when `sessionManager` is unavailable, and `createLauncher` no longer receiving `mpSessionId`.
- 35 existing full-payload `toEqual` assertions were updated, since the attribute is present on every call. The `#selectPlacements` `beforeEach` now pins a known `sessionManager`; previously these tests depended on an earlier test deleting `window.mParticle.sessionManager` without restoring it, which made results order-dependent.
- Test fixtures in `test/vitest.setup.ts` and `tests.spec.ts` moved from `getSession` to `getSessionId` to match the supported core surface.
- `dist/` is intentionally not included — the release workflow regenerates the bundle after merge.
- Additional testing worth doing: confirm on a long-lived page that a session which times out is reflected on the next `selectPlacements` call.

### DIFF
--- a/src/Rokt-Kit.ts
+++ b/src/Rokt-Kit.ts
@@ -965,7 +965,7 @@ class RoktKit implements KitInterface {
   }
 
   private readMpSessionId(): string | undefined {
-    const sessionManager = mp() && mp().sessionManager;
+    const sessionManager = mp()?.sessionManager;
     if (!sessionManager) {
       return undefined;
     }
--- a/src/Rokt-Kit.ts
+++ b/src/Rokt-Kit.ts
@@ -161,7 +161,7 @@ interface MParticleExtended {
   logEvent(name: string, type: number, attrs?: Record<string, unknown>): void;
   EventType: { Other: number };
   getInstance(): MParticleInstance;
-  sessionManager?: { getSession(): string };
+  sessionManager?: { getSession?(): string; getSessionId?(): string };
   _getActiveForwarders(): Array<{ name: string }>;
   config?: { isLocalLauncherEnabled?: boolean; isLoggingEnabled?: boolean };
   captureTiming?(metricName: string): void;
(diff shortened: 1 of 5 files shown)


src/Rokt-Kit.ts
```diff
@@ -963,20 +964,31 @@ class RoktKit implements KitInterface {
...
+      return undefined;
+    }
+
+    if (typeof sessionManager.getSessionId === 'function') {
+      return sessionManager.getSessionId() || undefined;
+    }
+
+    if (typeof sessionManager.getSession === 'function') {
+      return sessionManager.getSession() || undefined;
+    }
```
alexs-mparticle: `getSession()` is a deprecated method so we can leave it out.

src/Rokt-Kit.ts
```diff
@@ -963,20 +964,31 @@ class RoktKit implements KitInterface {
     }
   }
 
+  private readMpSessionId(): string | undefined {
+    const sessionManager = mp()?.sessionManager;
+    if (!sessionManager) {
+      return undefined;
+    }
+
+    if (typeof sessionManager.getSessionId === 'function') {
```
alexs-mparticle: Nit: We should create a `isFunction` util method.

alexs-mparticle (changes requested):

src/Rokt-Kit.ts
jamesnrokt: It ballooned the diff a bit but added

src/Rokt-Kit.ts
jamesnrokt: Was trying to keep the diff down but replaced all callers

alexs-mparticle (approved):

PR #119, opened 2026-08-17
fix: add reason codes to PAGE_VIEW_CAPTURE_FAILED logs

alexs-mparticle:
## Summary

Adds a `[reason: X]` suffix to all `PAGE_VIEW_CAPTURE_FAILED` log messages so Datadog can break down the failure modes without any behavioral changes.

| Reason | Meaning |
|--------|---------|
| `migration_retry` | Legacy `mpPageViews` key present but the migration write failed — currently causes infinite retries on every page load |
| `ls_unavailable` | localStorage is blocked (private browsing, user policy, Safari ITP) |
| `quota` | localStorage is accessible but the browser's hard quota is full |
| `exception` | Unexpected throw in the capture path unrelated to storage availability |

## Why before fixing

We have ~330/day errors from Belk alone. Before shipping behavioral changes we want to know which reason dominates — `ls_unavailable` and `migration_retry` have different fixes and different urgency.

## Changes

- `src/storage.ts` — adds `isLocalStorageAvailable()` probe (no caching — called only on failure paths)
- `src/pageViewStorage.ts` — appends `[reason: migration_retry]` to the migration failure message; re-exports `isLocalStorageAvailable`
- `src/Rokt-Kit.ts` — probes localStorage after each failure and appends the appropriate reason

## No behavior changes

All existing error paths fire identically — only the message string differs. No logic, no eviction, no early returns changed.

## Test plan

- [x] Lint pass, build pass
- [ ] After shipping, review logs for improved messaging

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -1,10 +1,11 @@
 import { isObject } from './utils';
 
+const LS_PROBE_KEY = '__rokt_ls_probe__';
+
 export function isLocalStorageAvailable(): boolean {
   try {
-    const probe = '__rokt_ls_probe__';
-    window.localStorage.setItem(probe, '1');
-    window.localStorage.removeItem(probe);
+    window.localStorage.setItem(LS_PROBE_KEY, '1');
+    window.localStorage.removeItem(LS_PROBE_KEY);
     return true;
   } catch {
     return false;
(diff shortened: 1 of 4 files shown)


src/storage.ts
```diff
@@ -1,5 +1,16 @@
 import { isObject } from './utils';
 
+export function isLocalStorageAvailable(): boolean {
+  try {
+    const probe = '__rokt_ls_probe__';
```
crisryantan: nit, outside the function so we dont redeclare every time

crisryantan (approved):

PR #122, opened 2026-08-18
feat: forward terminate from the kit to the Rokt launcher

jamesnrokt:
## Summary

Adds `RoktKit.terminate()`, which forwards to `launcher.terminate()` on the Rokt Web SDK.

This is the kit half of a supported teardown path for `window.mParticle.Rokt.terminate()`. The core SDK half is in mParticle/mparticle-web-sdk#1328; that PR's `RoktManager.terminate()` delegates to this method. **This PR should merge and release first** — the manager guards the call with `isFunction`, so until a kit carrying this change is on the page, `mParticle.Rokt.terminate()` logs that the attached kit version does not support it.

Partners currently tear down placements with the undocumented `window.Rokt.currentLauncher?.terminate()`. This makes the same call reachable through the documented interface.

### Behaviour

- **Ready:** returns `launcher.terminate()` directly, so callers can `await` the teardown. The launcher's own `terminate` races an internal 500ms timeout and never rejects.
- **Not ready:** logs `Rokt Kit: Not initialized` and resolves. Teardown is idempotent — with no launcher there is nothing to tear down.

### The launcher reference is deliberately left in place

`terminate()` does not null out `this.launcher` or `window.Rokt.currentLauncher`. The Rokt Web SDK memoizes a single launcher per page (`createLauncher` returns the cached promise once fulfilled), so clearing our references could not buy the caller a fresh launcher — it would only flip `isKitReady()` to false and leave every later `selectPlacements` call queued forever with no drain path.

Leaving state untouched makes this exactly equivalent to the `window.Rokt.currentLauncher.terminate()` call partners already make, just reachable through a supported API. There is a test pinning this.

## Testing Plan

`npm run lint`, `npm run build`, `npm run test` all pass — 313/313 Vitest tests, including 6 new `#terminate` cases covering:

- `launcher.terminate` is called when fully initialized
- the launcher's promise is returned and awaitable
- resolves as a no-op, without touching the launcher, when uninitialized
- resolves as a no-op when initialized but the launcher is missing
- launcher references survive teardown so the kit stays ready
- end-to-end through `init()` + `attachKit` in test mode

All 6 fail against `main` without the source change.

### Suggested manual check

On a page with a rendered placement, `await window.Rokt.currentLauncher.terminate()` and `await window.mParticle.Rokt.terminate()` (once mParticle/mparticle-web-sdk#1328 is also deployed) should be indistinguishable.

### DIFF
--- a/test/src/tests.spec.ts
+++ b/test/src/tests.spec.ts
@@ -216,6 +216,8 @@ describe('Rokt Forwarder', () => {
 
   afterEach(() => {
     (window as any).mParticle.forwarder.userAttributes = {};
+    (window as any).mParticle.forwarder._launcherTerminated = false;
+    (window as any).mParticle.forwarder._recreateLauncherPromise = null;
     delete (window as any).mParticle.forwarder.launcherOptions;
     delete (window as any).mParticle.Rokt.launcherOptions;
   });
@@ -3328,9 +3330,8 @@ describe('Rokt Forwarder', () => {
       expect(errorMessage).toBe('Rokt Kit: Not initialized');
     });
 
-    // The Rokt Web SDK memoizes one launcher per page, so terminate must not
-    // clear the kit's launcher references — doing so would flip the kit to
-    // not-ready with no way back.
+    // Leave the kit ready after terminate. A later createLauncher (SPA) can
+    // still mint a new instance because the Web SDK drops its memoized launcher.
     it('should leave the launcher references intact so the kit stays ready', async () => {
       const launcher = {
         terminate: () => Promise.resolve(),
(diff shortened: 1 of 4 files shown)


test/src/tests.spec.ts
```diff
@@ -3249,6 +3249,139 @@ describe('Rokt Forwarder', () => {
...
+        await expect((window as any).mParticle.forwarder.terminate()).resolves.toBeUndefined();
+      } finally {
+        window.console.error = originalConsoleError;
+      }
+
+      expect(errorMessage).toBe('Rokt Kit: Not initialized');
+    });
+
+    // The Rokt Web SDK memoizes one launcher per page, so terminate must not
+    // clear the kit's launcher references — doing so would flip the kit to
```
mattbodle, line 3332: Typically createLauncher can be called again after terminate, usually for SPAs.
Can this create another launcher instance in that flow and continue? (given the launcher reference is maintained)
Create a test to capture this scenario

crisryantan (changes requested):
Let's fix the types since we are already making changes

crisryantan (approved):

PR #125, opened 2026-08-19
fix: retry writePageViews with fewer records on quota failure

alexs-mparticle:
## Summary

- When `writePageViews` fails due to localStorage quota, evict the oldest record and retry — continuing until the write succeeds or no records remain
- Uses `let` + immutable `slice(1)` instead of mutating a `const` array
- Keeps the most-recent page views under quota pressure rather than discarding the entire write

## Why

~385 `quota` `PAGE_VIEW_CAPTURE_FAILED` errors/day in Datadog. The browser's total origin localStorage is full, so the write fails even with the 25-record cap in place. Evicting progressively fewer records may succeed when there is *some* space available, just not enough for all 25.

## Test plan

- [x] Existing: persists views and caps at 25 most-recent
- [x] New: single eviction — fails once, succeeds with 4 records (oldest dropped)
- [x] New: multi-eviction — fails 3 times, succeeds with 2 records
- [x] New: pre-seeded storage — 5 existing + 1 new, write of 6 fails, retries with 5 succeed (oldest of 6 dropped)
- [x] New: total failure — all writes fail, returns false
- [x] Lint and build pass

### DIFF
--- a/src/Rokt-Kit.ts
+++ b/src/Rokt-Kit.ts
@@ -884,7 +884,7 @@ class RoktKit implements KitInterface {
       const pageView = buildPageEvent(event);
       pageViews.push(pageView);
 
-      if (!writePageViews(pageViews)) {
+      if (writePageViews(pageViews) === 0) {
         const reason = isLocalStorageAvailable() ? 'quota' : 'ls_unavailable';
         this.loggingService?.log({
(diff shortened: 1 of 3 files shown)


alexs-mparticle:
closed in favor of #126

PR #129, opened 2026-08-20
feat: capture UTM params from page views and forward to selectPlacements

alexs-mparticle:
## Summary

- Adds `captureUtmParams()` to `pageViewStorage.ts` — reads the 5 standard UTM params (`utm_source`, `utm_medium`, `utm_campaign`, `utm_term`, `utm_content`) from `window.location.search` and stores them in localStorage under the `mp-rokt-kit` namespace (first-touch-wins)
- Forwards captured UTMs to Rokt as `page_view_attributes` on every `selectPlacements` call; field is omitted entirely when no UTMs were seen this session
- Clears stored UTMs on `SESSION_END`, consistent with how `pageViews` is cleared

## Design decisions

- **First touch wins**: UTMs are captured once per session and not overwritten, matching standard landing-page attribution behaviour
- **localStorage, not sessionStorage**: session boundary is the mParticle session (cleared on `SESSION_END`), not the browser tab — consistent with existing page view storage
- **Kit-only**: reads `window.location.search` directly; no core SDK changes required
- **`page_view_attributes` is a new contract**: backend wiring is a follow-up; extra attributes are ignored by Rokt until then

## Test plan

- [x] `captureUtmParams` stores present UTM keys on first call
- [x] Only stores keys that are present (partial set)
- [x] No-ops when URL has no UTM params
- [x] First touch wins — does not overwrite on subsequent calls
- [x] `loadUtmParams` returns `null` when nothing stored
- [x] `clearUtmParams` removes the field without affecting other namespace fields
- [x] Diagnostic logs emitted on capture success and write failure

### DIFF
--- a/src/Rokt-Kit.ts
+++ b/src/Rokt-Kit.ts
@@ -263,6 +263,7 @@ const USER_IDENTIFIED_IN_WORKSPACE_KEY = 'userIdentifiedInWorkspace';
 const MESSAGE_TYPE_PAGE_VIEW = 3; // mParticle MessageType.PageView
 const MESSAGE_TYPE_SESSION_END = 2; // mParticle MessageType.SessionEnd
 const PAGE_EVENTS_KEY = 'page_events';
+const PAGE_VIEW_ATTRIBUTES_KEY = 'page_view_attributes';
 const MPARTICLE_SESSION_ID_KEY = 'mparticle_session_id';
 
 // Bound on how long selectPlacements will wait for an in-flight Workspace
@@ -880,7 +881,6 @@ class RoktKit implements KitInterface {
 
     try {
       pageUrl = sanitizeUrl(window.location.href);
-      captureUtmParams(this.loggingService);
 
       const pageViews = loadPageViews(this.loggingService);
       const pageView = buildPageEvent(event);
@@ -1246,6 +1246,7 @@ class RoktKit implements KitInterface {
   public process(event: SDKEvent): string {
     if (!this.isTargetingDisabled()) {
       if (event.EventDataType === MESSAGE_TYPE_PAGE_VIEW) {
+        captureUtmParams(this.loggingService);
         this.capturePageView(event);
       }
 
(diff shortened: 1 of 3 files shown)


src/pageViewStorage.ts
```diff
@@ -92,6 +91,43 @@ export function buildPageEvents(pageViews: PageEvent[]): PageEvent[] {
...
+    const reason = isLocalStorageAvailable() ? 'quota' : 'ls_unavailable';
+    loggingService?.log({
+      message: `Rokt Kit: Failed to persist UTM params [reason: ${reason}]`,
+      code: 'UTM_CAPTURE_FAILED',
+    });
+    return;
+  }
+  loggingService?.log({
+    message: `Rokt Kit: Captured UTM params [${captured}]`,
+    code: 'UTM_CAPTURE_SUCCESS',
```
mattbodle: assuming this is just for the initial rollout verification and then we will remove?

src/Rokt-Kit.ts
```diff
@@ -1484,18 +1472,18 @@ class RoktKit implements KitInterface {
...
     const mpSessionId = this.readMpSessionId();
-    const mpDeviceId = this.readMpDeviceId();
 
     const selectPlacementsAttributes: Record<string, unknown> = {
       ...(filteredUserIdentities as Record<string, unknown>),
       ...filteredAttributes,
       ...optimizelyAttributes,
       ...sessionAttributes,
       ...(pageEvents.length ? { [PAGE_EVENTS_KEY]: JSON.stringify(pageEvents) } : {}),
+      ...(utmParams ? { page_view_attributes: utmParams } : {}),
```
mattbodle: should page_view_attributes be a const key like the others?

src/Rokt-Kit.ts
```diff
@@ -880,24 +880,18 @@ class RoktKit implements KitInterface {
 
     try {
       pageUrl = sanitizeUrl(window.location.href);
+      captureUtmParams(this.loggingService);
```
mattbodle: Should this sit inside capturePageView or maybe on line 1250?

src/pageViewStorage.ts
alexs-mparticle: Correct. I'll be removing these after the data has had time to soak.

jaissica12 (approved):

